PR #4, opened 2021-09-20
avoid incompatibility with future warnings hazard

matklad:
`#![deny(warnings)]` is an antipattern:

* it breaks downstream when compiler adds new warnings
* it has to be repeated for every crate (currenty, eg, tests don't deny warnings)
* it makes local development harder

A better approach is to deny warnings on CI


ParkMyCar:
makes sense to me, thanks for the PR!

PR #7, opened 2021-09-20
add inlined fast-path to drop

matklad:


matklad:
not 100% sure we need this, but in theory this seems nice.

epage:
I've played with outlining heap paths in `KString` (specifically, `clone`) but so far haven't seen any advantages.  However, I think this makes the code look nicer.

ParkMyCar:
Let me add some benchmarks around drop so we can make a more clear decision on this

ParkMyCar:
Thanks for the work here @matklad, sorry it took me so long to get back to this! I landed these changes in #78

PR #9, opened 2021-09-20
Remove potential bounds check from a hot path

matklad:


epage:
Huh, I tried this in `KString` and my benchmark for `s.is_empty()` saw no change.  No idea if impact is just that small or if I have a bug in the [benchmark](https://github.com/cobalt-org/kstring/blob/master/benches/access.rs).

ParkMyCar:
Similar to #7, let me add some bench marks here so we can make a better decision

matklad:
It might be hard to get a benchmark here: not-taken branches are “free” in terms of CPU time.

I would look at the generated assembly instead: with a bounds check, we should have a panic, and that might makes call-sites no-noexcept, and that might bloat code by creating extra unwinding landing pads.

matklad:
That is, the biggets win here should be making the function panic-free and allowing compiler to optimize call-sites based on that.

ParkMyCar:
Implemented in #114

Issue #12, opened 2021-09-20
Consider mutable representation for heap variant


matklad:
Currently, the heap variant is `Arc<str>`. This has a couple drawbacks: 

* Arc wastes code/space for weak pointer support which we don't use
* mutating the string requires  a full copy of data
* (need to re-check this) pointer in `Arc<str>` points to the start of the ArcInner, not to the start of str, so there's extra offset during deref.

An ideal representation would be to store `*const str`, which points:

```            
                             here
                               V
| rc: AtomicUsize | cap: usize | str byte data | spare capacity |
```

This should allow us to just mutate the string just fine (using `Arc::make_mut`-like API). 

Admittedly, this increases the complexity of the implementation quit a bit.

novacrazy:
How does one even allocate a struct like this?
```rust
#[repr(C)]
struct Inner {
    rc: AtomicUsize,
    cap: usize,
    data: str,
}
```
as I get to the point where `alloc::alloc(layout) as *mut Inner` but it's rejected due to casting a thin-pointer to a fat-pointer. And in fact, even how I define the layout is dubious. Three `usize`'s and `cap` bytes after? Can't determine it directly due to the entire struct being unsized now.

Otherwise, I have a little prototype for a version that doesn't point directly to a `str` in the middle of the struct: https://gist.github.com/novacrazy/6541c9000b12bed7f55edcbc0b4eea1b

Which does generate offset instructions on deref, unfortunately.

matklad:
>How does one even allocate a struct like this?

In today's Rust, via toil, pain, and suffering. 

One way is to make it just `struct Inner { rc: AtomicUsize, cap: usize }` and just maintain the invariant that whatever follows the strut in memory is `str`. 

If you want to keep it a proper DST you might want to look at https://github.com/rust-analyzer/rowan/blob/master/src/arc.rs which does a similar thing (the code originated in triomphe). 

@CAD97 also authored https://github.com/CAD97/pointer-utils which is a bunch of proper safe abstractions around these ideas.

NobodyXu:
@novacrazy You can take a look at [triomphe](https://github.com/Manishearth/triomphe), which povides a Arc and ThinArc that can store ad create DST easily and safely.

ParkMyCar:
I think this is a great idea, I'm wondering if we can store the ref count on the stack though, ideally make cloning even faster, e.g.

```
pub struct ArcStr {
    ref_count: AtomicUsize,
    pointer: *const u8,
[...]

PR #16, opened 2021-09-28
Add && Impl new trait ToCompactStr

NobodyXu:
Similar to [`std::string::ToString`](https://doc.rust-lang.org/std/string/trait.ToString.html), `ToCompactStr` convert any value implements `fmt::Display` into a `CompactStr` efficiently.

However, due to the fact that `Arc::<[MaybeUninit<T>]>` is still unstable, there is no way to implement `HeapString` efficiently without first creating a `String`, then copy it onto the buffer and deallocate the original `String`.

It seems that the compiler may not be able to optimize these calls, especially with `std::format!` calling `format`.

### DIFF
--- a/.github/workflows/msrv.yml
+++ b/.github/workflows/msrv.yml
@@ -19,10 +19,8 @@ jobs:
       - uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          toolchain: stable
-      - name: install cargo hack
-        run: cargo install cargo-hack
+          toolchain: 1.49
       - name: cargo test msrv..
         run: |
           cd compact_str
-          cargo hack test --feature-powerset --optional-deps --version-range 1.49..
+          cargo test --all-features
--- a/.github/workflows/msrv.yml
+++ b/.github/workflows/msrv.yml
@@ -19,9 +19,10 @@ jobs:
       - uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          default: true
-          toolchain: 1.49
+          toolchain: stable
+      - name: install cargo hack
+        run: cargo install cargo-hack
       - name: cargo test msrv..
         run: |
           cd compact_str
-          cargo test --all-features
+          cargo hack test --feature-powerset --optional-deps --version-range 1.49..
--- a/.github/workflows/msrv.yml
+++ b/.github/workflows/msrv.yml
@@ -19,6 +19,7 @@ jobs:
       - uses: actions-rs/toolchain@v1
         with:
           profile: minimal
+          default: true
           toolchain: 1.49
       - name: cargo test msrv..
         run: |
--- a/compact_str/Cargo.toml
+++ b/compact_str/Cargo.toml
@@ -12,12 +12,15 @@ keywords = ["string", "compact", "small", "memory", "mutable"]
 categories = ["encoding", "parsing", "memory-management", "text-processing"]
 
 [features]
+to-compact-str-float-spec = ["ryu", "to-compact-str-specialisation"]
 to-compact-str-int-spec = ["itoa", "to-compact-str-specialisation"]
 to-compact-str-specialisation = ["castaway"]
 
 [dependencies]
 bytes = { version = "1", optional = true }
 serde = { version = "1", optional = true }
+
+# Specialisation
 castaway = { version = "0.2", optional = true }
 itoa = { version = "1", optional = true }
 ryu = { version = "1", optional = true }
--- a/compact_str/Cargo.toml
+++ b/compact_str/Cargo.toml
@@ -16,6 +16,7 @@ categories = ["encoding", "parsing", "memory-management", "text-processing"]
 [dependencies]
 bytes = { version = "1", optional = true }
 serde = { version = "1", optional = true }
+castaway = { version = "0.2", optional = true }
 
 [dev-dependencies]
 proptest = { version = "1", default-features = false, features = ["std"] }
--- a/compact_str/Cargo.toml
+++ b/compact_str/Cargo.toml
@@ -12,7 +12,7 @@ keywords = ["string", "compact", "small", "memory", "mutable"]
 categories = ["encoding", "parsing", "memory-management", "text-processing"]
 
 [features]
-to-compact-str-int-spec = ["to-compact-str-specialisation"]
+to-compact-str-int-spec = ["itoa", "to-compact-str-specialisation"]
 to-compact-str-specialisation = ["castaway"]
 
 [dependencies]
--- a/compact_str/Cargo.toml
+++ b/compact_str/Cargo.toml
@@ -12,6 +12,7 @@ keywords = ["string", "compact", "small", "memory", "mutable"]
 categories = ["encoding", "parsing", "memory-management", "text-processing"]
 
 [features]
+to-compact-str-int-spec = ["to-compact-str-specialisation"]
 to-compact-str-specialisation = ["castaway"]
 
 [dependencies]
--- a/compact_str/Cargo.toml
+++ b/compact_str/Cargo.toml
@@ -11,19 +11,13 @@ readme = "../README.md"
 keywords = ["string", "compact", "small", "memory", "mutable"]
 categories = ["encoding", "parsing", "memory-management", "text-processing"]
 
-[features]
-to-compact-str-float-spec = ["ryu", "to-compact-str-specialisation"]
-to-compact-str-int-spec = ["itoa", "to-compact-str-specialisation"]
-to-compact-str-specialisation = ["castaway"]
-
 [dependencies]
 bytes = { version = "1", optional = true }
 serde = { version = "1", optional = true }
 
-# Specialisation
-castaway = { version = "0.2.1", optional = true }
-itoa = { version = "1", optional = true }
-ryu = { version = "1", optional = true }
+castaway = "0.2.1"
+itoa = "1"
+ryu = "1"
 
 [dev-dependencies]
 proptest = { version = "1", default-features = false, features = ["std"] }
--- a/compact_str/Cargo.toml
+++ b/compact_str/Cargo.toml
@@ -11,7 +11,8 @@ readme = "../README.md"
 keywords = ["string", "compact", "small", "memory", "mutable"]
 categories = ["encoding", "parsing", "memory-management", "text-processing"]
 
-# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+[features]
+to-compact-str-specialisation = ["castaway"]
 
 [dependencies]
 bytes = { version = "1", optional = true }
--- a/compact_str/Cargo.toml
+++ b/compact_str/Cargo.toml
@@ -21,7 +21,7 @@ bytes = { version = "1", optional = true }
 serde = { version = "1", optional = true }
 
 # Specialisation
-castaway = { version = "0.2", optional = true }
+castaway = { version = "0.2.1", optional = true }
 itoa = { version = "1", optional = true }
 ryu = { version = "1", optional = true }
 
--- a/compact_str/Cargo.toml
+++ b/compact_str/Cargo.toml
@@ -18,6 +18,7 @@ to-compact-str-specialisation = ["castaway"]
 bytes = { version = "1", optional = true }
 serde = { version = "1", optional = true }
 castaway = { version = "0.2", optional = true }
+itoa = { version = "1", optional = true }
 
 [dev-dependencies]
 proptest = { version = "1", default-features = false, features = ["std"] }
--- a/compact_str/Cargo.toml
+++ b/compact_str/Cargo.toml
@@ -20,6 +20,7 @@ bytes = { version = "1", optional = true }
 serde = { version = "1", optional = true }
 castaway = { version = "0.2", optional = true }
 itoa = { version = "1", optional = true }
+ryu = { version = "1", optional = true }
 
 [dev-dependencies]
 proptest = { version = "1", default-features = false, features = ["std"] }
--- a/compact_str/src/lib.rs
+++ b/compact_str/src/lib.rs
@@ -698,8 +698,10 @@ pub trait ToCompactStr {
     /// Basic usage:
     ///
     /// ```
+    /// use compact_str::ToCompactStr;
+    ///
     /// let i = 5;
-    /// let five = compact_str::CompactStr::from("5");
+    /// let five = "5".to_compact_str();
     ///
     /// assert_eq!(five, i.to_string());
     /// ```
--- a/compact_str/src/lib.rs
+++ b/compact_str/src/lib.rs
@@ -16,20 +16,15 @@
 use core::borrow::Borrow;
 use core::cmp::Ordering;
 use core::fmt;
-use core::hash::{
-    Hash,
-    Hasher,
-};
+use core::hash::{Hash, Hasher};
 use core::iter::FromIterator;
 use core::ops::Deref;
-use core::str::{
-    FromStr,
-    Utf8Error,
-};
+use core::str::{FromStr, Utf8Error};
 use std::borrow::Cow;
 
 mod asserts;
 mod features;
+mod utility;
 
 mod repr;
 use repr::Repr;
@@ -686,3 +681,44 @@ impl fmt::Write for CompactStr {
 }
 
 crate::asserts::assert_size_eq!(CompactStr, String);
+
+/// A trait for converting a value to a `CompactStr`.
+///
+/// This trait is automatically implemented for any type which implements the
+/// [`Display`] trait. As such, `ToCompactStr` shouldn't be implemented directly:
+/// [`Display`] should be implemented instead, and you get the `ToCompactStr`
+/// implementation for free.
+///
+/// [`Display`]: fmt::Display
+pub trait ToCompactStr {
+    /// Converts the given value to a `CompactStr`.
+    ///
+    /// # Examples
+    ///
+    /// Basic usage:
+    ///
+    /// ```
+    /// let i = 5;
+    /// let five = compact_str::CompactStr::from("5");
+    ///
+    /// assert_eq!(five, i.to_string());
+    /// ```
+    fn to_compact_str(&self) -> CompactStr;
+}
+
+/// # Panics
+///
+/// In this implementation, the `to_compact_str` method panics
+/// if the `Display` implementation returns an error.
+/// This indicates an incorrect `Display` implementation
+/// since `std::fmt::Write for String` never returns an error itself and
+/// the implementation of `ToCompactStr::to_compact_str` only panic if the `Display`
+/// implementation is incorrect..
+impl<T: fmt::Display + ?Sized> ToCompactStr for T {
+    #[inline]
+    fn to_compact_str(&self) -> CompactStr {
+        CompactStr {
+            repr: Repr::from_fmt(self),
+        }
+    }
+}
--- a/compact_str/src/lib.rs
+++ b/compact_str/src/lib.rs
@@ -18,8 +18,14 @@ use core::cmp::Ordering;
 use core::fmt;
 use core::hash::{Hash, Hasher};
 use core::iter::FromIterator;
-use core::ops::Deref;
-use core::str::{FromStr, Utf8Error};
+use core::ops::{
+    Add,
+    Deref,
+};
+use core::str::{
+    FromStr,
+    Utf8Error,
+};
 use std::borrow::Cow;
 
 mod asserts;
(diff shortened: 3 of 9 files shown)


ParkMyCar:
Thanks for the PR!

I really like the idea of a `ToCompactStr` trait. But you're right, without the nightly compiler we can't do this very efficiently. The `HeapString` representation will be changing soon, see #12, I plan to spend some time on it this weekend. Let's hold off on landing this until we make the changes to `HeapString`

NobodyXu:
Sure, I will wait for #12

ParkMyCar:
Hey @NobodyXu, now that `CompactStr` is more stable, would you want to take another shot at this?

I think the idea of a `ToCompactStr` trait is great, I'm a bit apprehensive of manually implementing it for any type that implements `Display` though because of performance. Maybe we could define the trait, and then manually implement for a few common types like all the number types (e.g. `u8`, `isize`, `f32`, etc), and then maybe `bool` and `char`? IMO that would go a fairly long way for common use cases, what do you think?

For efficiently converting number types to strings take a look at [`itoa`](https://github.com/dtolnay/itoa) and [`ryu`](https://github.com/dtolnay/ryu)

NobodyXu:
> Hey @NobodyXu, now that `CompactStr` is more stable, would you want to take another shot at this?

Yes, sure.

> I think the idea of a `ToCompactStr` trait is great, I'm a bit apprehensive of manually implementing it for any type that implements `Display` though because of performance. Maybe we could define the trait, and then manually implement for a few common types like all the number types (e.g. `u8`, `isize`, `f32`, etc), and then maybe `bool` and `char`? IMO that would go a fairly long way for common use cases, what do you think?

Hmmm, I don't like the idea of implementing only for common types, as it would make `ToCompactStr` less useful and
provide different semantics compared to `ToString`.

For common number types, maybe we can provide a optional specialization for them using
 [castaway](https://crates.io/crates/castaway) and gate it using a feature flag?

ParkMyCar:
> Hmmm, I don't like the idea of implementing only for common types, as it would make ToCompactStr less useful and
provide different semantics compared to ToString.
Fair enough, that makes sense to me!

Ugh I wish [specialization](https://rust-lang.github.io/rfcs/1210-impl-specialization.html) was stable. `castaway` looks pretty cool, let's give it a shot!

NobodyXu:
> > Hmmm, I don't like the idea of implementing only for common types, as it would make ToCompactStr less useful and
> > provide different semantics compared to ToString.
> 
> Fair enough, that makes sense to me!
> 
> Ugh I wish [specialization](https://rust-lang.github.io/rfcs/1210-impl-specialization.html) was stable. Regardless, `castaway` looks pretty cool, let's give it a shot!

I am quite busy recently, so it might take me some time to implement it.

NobodyXu:
Sorry I accidentally closed this PR, now it is reopened.

NobodyXu:
@ParkMyCar I have implemented trait `ToCompactStr` without specialisation using [castaway](https://docs.rs/castaway/0.1.2/castaway/macro.cast.html).

I will implement the specialisation later, but now you can probably start looking at my code and providing me with feedback so that we can ensure the implementation is correct before doing any optimization.

compact_str/src/lib.rs
```diff
@@ -16,20 +16,15 @@
 use core::borrow::Borrow;
 use core::cmp::Ordering;
 use core::fmt;
-use core::hash::{
-    Hash,
-    Hasher,
-};
+use core::hash::{Hash, Hasher};
```
ParkMyCar, line 19: Can you re-run `cargo +nightly fmt`? We use some rules which are nightly only

compact_str/src/lib.rs
```diff
@@ -686,3 +681,44 @@ impl fmt::Write for CompactStr {
...
+    /// # Examples
+    ///
+    /// Basic usage:
+    ///
+    /// ```
+    /// let i = 5;
+    /// let five = compact_str::CompactStr::from("5");
+    ///
+    /// assert_eq!(five, i.to_string());
+    /// ```
```
ParkMyCar, line 705: This example doesn't actually use `.to_compact_str()`?

compact_str/src/lib.rs
```diff
@@ -686,3 +681,44 @@ impl fmt::Write for CompactStr {
...
+/// This indicates an incorrect `Display` implementation
+/// since `std::fmt::Write for String` never returns an error itself and
+/// the implementation of `ToCompactStr::to_compact_str` only panic if the `Display`
+/// implementation is incorrect..
+impl<T: fmt::Display + ?Sized> ToCompactStr for T {
+    #[inline]
+    fn to_compact_str(&self) -> CompactStr {
+        CompactStr {
+            repr: Repr::from_fmt(self),
+        }
```
ParkMyCar, line 722: `CompactStr` now implements `fmt::Write`, so I think this entire impl could just be
```
fn to_compact_str(&self) -> CompactStr {
  let mut compact = CompactStr::new("");
  write!(compact, "{}", self).expect("fmt::Display incorrectly implemented!");
  compact
}
```

compact_str/src/utility.rs
```diff
@@ -0,0 +1,58 @@
...
+/// A special kind of sink that records the size of bytes writen into it.
+#[derive(Debug, Default)]
+pub struct Sink(pub usize);
+impl Sink {
+    pub fn count(args: fmt::Arguments) -> usize {
+        let mut sink = Sink::default();
+        write!(&mut sink, "{}", args).unwrap();
+        sink.0
+    }
+}
```
ParkMyCar, line 13: This is pretty neat! See above though, I don't think we need to do this since `CompactStr` impls `fmt::Write`

ParkMyCar (changes requested):
Thanks for doing this work!

compact_str/src/lib.rs
NobodyXu: Oops, I simply copied it from the old PR diff.
I will change that.

compact_str/src/lib.rs
NobodyXu: That would make the code much shorter.

However, I think we should benchmark the code first, as the current implementation might provide better performance than using fmt::Write, since everything is written in one-go.

For size <= MAX_SIZE, current implementation can be optimized into one copy of array, or no copy at all.

For size > MAX_SIZE, current implementation would only allocate once, where fmt::Write could allocate multiple times.

NobodyXu: I have pushed a new commit that optimize the case where size > `MAX_SIZE` and preallocate the `String` with the exact size and use `fmt::Write` to write the format string.

This would avoid the repeated allocation that occurs if using `fmt::Write` implementation of `CompactStr`.

compact_str/src/lib.rs
NobodyXu: Done.

compact_str/src/lib.rs
NobodyXu: That explains why the CI complains about the code style, while my neovim is configured to automatically format my code using `cargo fmt`.

I will run `cargo +nightly fmt` just before merging so that I don't have to run it repeatively.

compact_str/src/repr/mod.rs
```diff
@@ -108,6 +108,23 @@ impl Repr {
...
+    #[inline]
+    pub fn from_fmt(val: impl fmt::Display) -> Self {
+        let len = crate::count!("{}", val);
+
+        if len == 0 {
+            EMPTY
+        } else if len <= MAX_SIZE {
+            let mut buffer = [0_u8; MAX_SIZE];
+            let text = format_into_buffer(&mut buffer, &val);
+            Self::new(text)
```
ParkMyCar, line 120: We should either `impl Write` for `Repr` or have this method return a `CompactStr`, for two reasons:
1. Writing into a buffer and then using that buffer to create a `Repr` results in an additional copy
2. Using something like `format_into_buffer(...)` is another part of code that manually writing bytes and ideally we have as few of those as possible, since these are the easiest things to get wrong

ParkMyCar, line 120: We should either `impl Write` for `Repr` or have this method return a `CompactStr`, for two reasons:
1. Writing into a buffer and then using that buffer to create a `Repr` results in an additional copy
2. Using something like `format_into_buffer(...)` is another part of code that manually writing bytes and ideally we have as few of those as possible, since these are the easiest things to get wrong

compact_str/src/lib.rs
ParkMyCar: I ran some benchmarks with this implementation and with just using `write!(...)`. In the simple case (e.g. `&str`, `usize`, etc.) just using `write!(...)` is a lot faster, especially for short strings. It's when we have long strings with multiple arguments (e.g. a struct) that counting all of the bytes beforehand and pre-allocating saves a lot of time.

Given this, the approach of initially counting bytes works for me, with the idea that we can optimize for the simple cases later on.

ParkMyCar: I ran some benchmarks with this implementation and with just using `write!(...)`. In the simple case (e.g. `&str`, `usize`, etc.) just using `write!(...)` is a lot faster, especially for short strings. It's when we have long strings with multiple arguments (e.g. a struct) that counting all of the bytes beforehand and pre-allocating saves a lot of time.

Given this, the approach of initially counting bytes works for me, with the idea that we can optimize for the simple cases later on.

compact_str/src/repr/mod.rs
NobodyXu: I have fixed this performance issue by implementing `InlineString::from_fmt` and call the `format_into_buffer` directly on the buffer used by `InlineString`.

NobodyXu: I have now refactored the code and inline `format_into_buffer` into `InlineString::from_fmt`.

Since `format_into_buffer` simply uses `std::io::Cursor`, it's unlikely to have any bugs.

NobodyXu: I have pushed another commit to simplify `InlineString::from_fmt`, by removing use of `std::io::Cursor`.

NobodyXu:
> For efficiently converting number types to strings take a look at [`itoa`](https://github.com/dtolnay/itoa) and [`ryu`](https://github.com/dtolnay/ryu)

I am doubtful on using [`ryu`](https://github.com/dtolnay/ryu) as it might give different output compared to the builtin `format!`, I think this might break some code.

NobodyXu:
> > Hmmm, I don't like the idea of implementing only for common types, as it would make ToCompactStr less useful and
> > provide different semantics compared to ToString.
> 
> Fair enough, that makes sense to me!
> 
> Ugh I wish [specialization](https://rust-lang.github.io/rfcs/1210-impl-specialization.html) was stable. Regardless, `castaway` looks pretty cool, let's give it a shot!

I just realized that `castaway` requires the type to be of `'static` lifetime, which is impossible to use within `ToCompactStr`.

I am now thinking of adding another trait `ToCompactStrFast` for the builtin type.

NobodyXu:
I also asked the question https://github.com/sagebind/castaway/issues/5 on `castaway` to see if there is any alternative solution.

NobodyXu:
> > > Hmmm, I don't like the idea of implementing only for common types, as it would make ToCompactStr less useful and
> > > provide different semantics compared to ToString.
> > 
> > 
> > Fair enough, that makes sense to me!
> > Ugh I wish [specialization](https://rust-lang.github.io/rfcs/1210-impl-specialization.html) was stable. Regardless, `castaway` looks pretty cool, let's give it a shot!
> 
> I just realized that `castaway` requires the type to be of `'static` lifetime due to `std::any::TypeId::of` requires static lifetime, which is impossible to use within `ToCompactStr`.
> 
> I am now thinking of adding another trait `FromBuiltinTypes`.
> 
> It is just unfortunate that Rust doesn't have specialisation stabilised.

With https://github.com/sagebind/castaway/pull/6 merged in and released as https://github.com/sagebind/castaway/releases/tag/0.2.0, this is no longer an issue.

NobodyXu:
@ParkMyCar I've added int and float specialisation, but there are two problems:
 - [ ] First is `castaway::LifetimeFree` is not implemented for `NonZeroUsize` so that we cannot provide specialisation for it yet https://github.com/sagebind/castaway/pull/7
 - [ ] Second is that test on float specialisation (line 421 of `tests.rs`) fails on power pc little endian 64-bit:
        Expected: 3.2
        Actual: 0.0
        It failed only after float specialisation is added.

compact_str/src/to_compact_str_specialisation.rs
```diff
@@ -0,0 +1,166 @@
...
+    #[cfg(feature = "to-compact-str-float-spec")]
+    // FIXME: The tests for float specialisation fail
+    // on power pc little endian 64-bit:
+    //
+    //     assert_eq!("3.2", 3.2_f32.to_compact_str());
+    //
+    // Expected: 3.2
+    // Actual: 0.0
+    //
+    // It failed only after float specialisation is added.
```
sagebind, line 28: Hey, just snooping on this PR as I like to see how Castaway is being used! I'll double-check to make sure that this bug isn't coming from Castaway's side.

NobodyXu: Awesome!

Since `castaway` is working as intended, I think the bug is more likely to be from the float formatting algorithm in std and ryu.

Seems like they had a hard time running on `powerpc64le` target.

sagebind: Yeah I figured that was the case, but I am beefing up my tests just to be sure. I'm confident in Castaway's logic, but when internally I have a `transmute_unchecked` function because the already-dangerous builtin `transmute` is _too strict_ you can't be too careful. 😀

NobodyXu:
I just tried to reproduce this on my own x86_64 linux box by running:

```
cross +nightly test --release --all-features --manifest-path=compact_str/Cargo.toml --target powerpc-unknown-linux-gnu
```

however, all tests passed without problem...

compact_str/src/to_compact_str_specialisation.rs
NobodyXu: This is confirmed to be a bug in rust as I can reproduce this on my own x86_64 linux box using `cross`, so would have to wait for rust to fix it.

.github/workflows/msrv.yml
```diff
@@ -19,10 +19,9 @@ jobs:
...
           profile: minimal
-          toolchain: stable
-      - name: install cargo hack
-        run: cargo install cargo-hack
+          default: true
+          toolchain: 1.49
       - name: cargo test msrv..
         run: |
           cd compact_str
-          cargo hack test --feature-powerset --optional-deps --version-range 1.49..
```
ParkMyCar, line 28: Why did you get rid of `cargo hack`? It's quite useful in making sure `compact_str` works for all versions of rustc since the MSRV, and all different combination of features?

compact_str/Cargo.toml
```diff
@@ -11,12 +11,20 @@ readme = "../README.md"
 keywords = ["string", "compact", "small", "memory", "mutable"]
 categories = ["encoding", "parsing", "memory-management", "text-processing"]
 
-# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+[features]
+to-compact-str-float-spec = ["ryu", "to-compact-str-specialisation"]
+to-compact-str-int-spec = ["itoa", "to-compact-str-specialisation"]
+to-compact-str-specialisation = ["castaway"]
```
ParkMyCar, line 17: I would really prefer to not have so many features, especially ones tied to performance, can we make these dependencies required instead of optional?

compact_str/src/repr/inline.rs
```diff
@@ -29,6 +32,25 @@ impl InlineString {
...
+
+        // format the val into buffer.
+        //
+        // note: in the case where len == MAX_SIZE, we'll overwrite the len, but that's okay because
+        // when reading the length we can detect that the last byte is part of UTF-8 and return a
+        // length of MAX_SIZE
+        write!(&mut buffer[..], "{}", val).expect("fmt::Display incorrectly implemented!");
+
+        InlineString { buffer }
+    }
```
ParkMyCar, line 52: This introduces another part of the code base that directly writes into the buffer and needs to know about the underlying bit layout, I'd like to prevent that. 

See my comment on `impl<T: fmt::Display> ToCompactStr for T` on how we can eliminate this

compact_str/src/lib.rs
```diff
@@ -737,3 +732,51 @@ impl Add<CompactStr> for String {
...
+impl<T: fmt::Display> ToCompactStr for T {
+    #[inline]
+    fn to_compact_str(&self) -> CompactStr {
+        #[cfg(feature = "to-compact-str-specialisation")]
+        if let Some(compact_str) = to_compact_str_specialisation::to_compact_str_specialised(self) {
+            return compact_str;
+        }
+
+        CompactStr {
+            repr: Repr::from_fmt(self),
```
ParkMyCar, line 779: Can we implement `fmt::Write` for `Repr`, and then here do something like:
```
let num_bytes = count(&self);
let mut repr = Repr::with_capacity(num_bytes);
write!(&mut repr, "{}", self).expect(...);
```

compact_str/src/tests.rs
```diff
@@ -365,3 +366,87 @@ fn test_plus_operator() {
...
+        ToCompactStr::to_compact_str(
+            &core::format_args!(
+                $fmt,
+                $(
+                    $args,
+                )*
+            )
+        )
+    };
+}
```
ParkMyCar, line 381: Is the purpose of this macro to be similar to `format!(...)`? If so, could we rename this to `format_compact!(...)`

compact_str/src/to_compact_str_specialisation.rs
```diff
@@ -0,0 +1,164 @@
...
+///  - CompactStr does not contain any lifetime
+///  - CompactStr is 'static
+///  - CompactStr is a container to `u8`, which is `LifetimeFree`.
+unsafe impl LifetimeFree for CompactStr {}
+
+const TRUE_COMPACT_STR: CompactStr = CompactStr::new_inline("true");
+const FALSE_COMPACT_STR: CompactStr = CompactStr::new_inline("false");
+
+#[inline(always)]
+pub(super) fn to_compact_str_specialised<T>(val: &T) -> Option<CompactStr> {
```
ParkMyCar, line 16: I'll need to take a closer look at this to figure out how it's actually zero cost, right now there seem to be some runtime checks?

Out of curiosity, is it not possible to do autoref specialization here?

ParkMyCar (changes requested):
Thanks for all the work here @NobodyXu! I've been pretty busy lately so haven't had a chance to look super in-depth, but I think we're getting close!

compact_str/src/to_compact_str_specialisation.rs
NobodyXu: According to @sagebind , all of these cast! can be optimized out by compiler.

I briefly read through the code and I also think it is possible for compiler to prune all the unused specialisation.

compact_str/Cargo.toml
NobodyXu: sure

.github/workflows/msrv.yml
NobodyXu: They are taking too long to run, probably because of the new features I added

compact_str/src/to_compact_str_specialisation.rs
sagebind: "runtime checks" is a bit hard to define in light of optimization; technically yes, Castaway macros generate program code to make appropriate checks. But these checks are trivial from the compiler's perspective (think `if size_of::<A>() == size_of::<B>()`) and are always reduced to a simple `true` or `false` at even just a basic optimization level; once this happens it is also trivial for the compiler to eliminate a branch in _your_ code, since it effectively looks like `if true {} else {}` (or `if false {} else {}` depending on which type it is called with) to rustc by the time that stage of compilation is reached.

> Out of curiosity, is it not possible to do autoref specialization here?

Nope, auto(de)ref techniques don't work inside a generic context, because ref specialization is resolved _before_ monomorphization happens. (Which is a good thing by the way, otherwise you'd have C++ problems, such as compilation failing "sometimes" depending on which concrete types you supply to a template.)

compact_str/Cargo.toml
NobodyXu: Done!

compact_str/src/tests.rs
NobodyXu: Done!

compact_str/src/repr/inline.rs
NobodyXu: Done!

compact_str/src/lib.rs
NobodyXu: Done!

I've taken things even further, by using `CompactStr::with_capacity` then use `fmt::Write` implementation for `CompactStr` to format `self` into `compact_str.`

.github/workflows/msrv.yml
NobodyXu: OK, I've reverted the changes to this workflow.

compact_str/src/lib.rs
NobodyXu: I've applied `cargo +nightly fmt --all`.

SInce this PR is pretty close to being merged, I would apply `cargo +nightly fmt --all` everytime I push new commits.

compact_str/src/to_compact_str_specialisation.rs
NobodyXu: @ParkMyCar [Here is the code](https://github.com/sagebind/castaway/blob/52987d6ebbdfcc451fb24585dcb34590f6e62657/src/utils.rs#L29) responsible for checking whether two non-`' static` types are the same, which is at the core of `castaway`.

I am pretty sure they can be easily evaluated at compile time.

ParkMyCar: > since it effectively looks like `if true {} else {}` (or `if false {} else {}`)
That's wild, super cool work @sagebind!

ParkMyCar (approved):
Thanks for all the hard work here @NobodyXu and making all the changes I asked for! 

Looks like there's just one failing test on PowerPC Little Endian with float formatting? If we can fix the test that would be great, otherwise I'm fine with ignoring it. Once you make that change we can merge!

Thanks again for all the work! 🙂

NobodyXu:
> Looks like there's just one failing test on PowerPC Little Endian with float formatting? If we can fix the test that would be great, otherwise I'm fine with ignoring it. Once you make that change we can merge!

Unfortunately that isna compiler bug https://github.com/rust-lang/rust/issues/96306 which I cannot fix.

ParkMyCar:
Merging, thanks @NobodyXu!

NobodyXu:
@ParkMyCar Now that we have more unit tests and fuzzing added for `ToCompactStr`, maybe it is a good time to release a new minor version?

PR #20, opened 2021-10-29
impl FromStr for CompactStr

### LINKED ISSUE #18: Consider impl trait FromStr for CompactStr
I got error when using StructOpts

```
    | |__________________________^ the trait `FromStr` is not implemented for `CompactStr`
```

ParkMyCar:
Fixes #18

This PR implements the `std::str::FromStr` trait for `CompactStr`, where the error type is `core::convert::Infallible`. The ergonomics here aren't _great_ because despite making the error type "infallible" we still have to call unwrap. Once the never type, `!`, is stabilized, we can use that instead.

### DIFF
--- a/compact_str/src/lib.rs
+++ b/compact_str/src/lib.rs
@@ -17,6 +17,7 @@ use core::{
     fmt,
     hash::{Hash, Hasher},
     ops::Deref,
+    str::FromStr,
 };
 
 mod repr;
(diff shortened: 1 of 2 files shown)


ParkMyCar:
Test failure on `X-Plat / macOS` seemed to be an infra issue. Trying again [here](https://github.com/ParkMyCar/compact_str/actions/runs/1400524524)

PR #22, opened 2021-10-31
draft: Optimize Option<CompactStr> to be the same size as CompactStr

### LINKED ISSUE #19: Is it possible to make Option<CompactStr> same size as Option<String>?
It would be very beneficial for me to reduce size of `Option<CompactStr>`. I use lots of it in my projects.

I made some tests, and it looks like `Option<ComactStr>` takes additional 8 bytes in compaction to `Option<String>`.

Rust has null pointer optimization (see https://doc.rust-lang.org/std/option/index.html#representation) but I not sure if it can by used in userland crate.

sizeofs of popular Strings.
```
String: 24 bytes
Option<String>: 24 bytes
CompactStr: 24 bytes
Option<CompactStr>: 32 bytes
KString: 24 bytes
Option<KString>: 24 bytes
SmartString<Compact>: 24 bytes
Option<SmartString<Compact>>: 32 bytes
SmolStr: 24 bytes
Option<SmolStr>: 24 bytes
```

ParkMyCar:
This is a proof of concept to show that it's possible in stable Rust to get `std::mem::size_of<CompactStr> == std::mem::size_of::<Option<CompactStr>>()`. 

Unfortunately as-is the performance of `CompactStr` is degraded. We would need to fix this before merging.

Fixes #19


ParkMyCar:
I'm pretty attached to the `Packed` representation. Specifically because that allows you to inline 12 ASCII character strings on 32-bit machines, and 12 characters is the most common length for filenames because of the 8.3 notation from the early days of FAT, e.g. `DSC_1234.jpg`.

That being said, maybe we can use the layout you proposed and only transmute for the `Packed` repr. In the common case this removes the transmute and thus the extra copy, at the cost of code complexity 🤔

ParkMyCar:
Completed in #105

PR #30, opened 2021-12-08
feature: Create a new heap variant to replace Arc<str>

ParkMyCar:
This PR changes our `HeapString` implementation from using an `Arc<str>` to a custom defined `ArcString` type. This type is more analogous to `Arc<[u8]>`, specifically it allocates a buffer which may be larger than the "string" it contains. This gets us closer to being a true drop-in replacement for `String` and allows us to write APIs for mutating a `CompactStr`


NobodyXu:
I think it’s totally unnecessary to reinvent Arc here.

We can use [slice-dst](https://docs.rs/slice-dst/latest/slice_dst/) to create an unsized Arc.

Or, we can use [triomphe](https://github.com/Manishearth/triomphe) which provides similar functionality but easier to use, and I had my PR relating to MaybeUninit usage merged in it.

ParkMyCar:
Thanks for the feedback @NobodyXu!

I did look at `tripomphe` and I actually saw your recent PR for adding an API for `MaybeUninit<T>`, but unfortunately it doesn't seem to have been released yet? Otherwise I played around with `trimophe::ThinArc` a bit but it doesn't provide an API to modify the underlying buffer, which is one of the goals of this change, so we can get closer to being a truly drop-in replacement for `String`.

NobodyXu:
> an API to modify the underlying buffer

Is that for the copy-on-write usage?

Well then, I think this does require creation of another unique `Arc` type, though I think it might be better to implement this in another crate in a generic way.

But then, that can be postponed until this feature is completed in `compact_str`.

This is quite a novel way to implement `CompactStr` to save memory while permit modification.

Thanks for taking your time to answer my question!

PR #44, opened 2022-01-14
perf: Lazily instantiate ref_count on ArcString

ParkMyCar:
This PR introduces a small perf improvement when creating the heap variant of a `CompactStr`. 

Currently when creating an `ArcString` we initialize the `ref_count` with `AtomicUsize::new(1)`, indicating there is one reference to the string buffer. This isn't great though because now users who never Clone the string, still have to pay the price of creating the `AtomicUsize`, and checking the value of it when modifying or dropping the string. This violates Rust's informal principle of "you shouldn't need to pay for something, if you don't use it".

To fix this, we make the `ref_count` an `Option<AtomicUsize>`, where `None` indicates there is one outstanding reference. In code paths where we have to check the ref count, this allows us to first check if `ref_count` is `None`, which prevents the need to perform atomic operations. When clone-d for the first time, we then instantiate the ref_count to `Some(AtomicUsize::new(2))`, and from there we use the `ref_count` as one normally would. 
Note, when decrementing the `ref_count` from 2 to 1, we do not reset back to `None` because another thread could also be concurrently decrementing the `ref_count` and there is no way to synchronize setting the `None` value.

### DIFF
--- a/compact_str/src/repr/heap/arc.rs
+++ b/compact_str/src/repr/heap/arc.rs
@@ -68,24 +68,17 @@ impl ArcString {
 
     #[inline]
     pub unsafe fn make_mut_slice(&mut self) -> &mut [u8] {
-        if self
-            .inner()
-            .ref_count
-            .compare_exchange(1, 0, Ordering::Acquire, Ordering::Relaxed)
-            .is_err()
-        {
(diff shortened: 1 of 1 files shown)


ParkMyCar:
jk, I realized this doesn't work with scoped threads

PR #47, opened 2022-01-24
feature: BoxString

ParkMyCar:
This PR adds another heap implementation called `BoxString`. It's a non-reference counted heap string that stores the capacity and string buffer on the heap. Eventually this will be the default implementation for heap storage, and users can opt into atomic reference counting (the current impl) via a new type.


ParkMyCar:
bleh, I messed something up with git branches, this is now a duplicate of #56

PR #62, opened 2022-02-28
impl From<Cow<'_, str>> for CompactStr

mcronce:
Super small; just something I noticed when I went to benchmark replacing `smartstring` with `compact_str` in one of my projects :)

Kudos on the awesome crate!  It turns out that, at least in `imdb_async`, I deal with a _lot_ of strings that are 24 bytes in length.  Average performance seems to be approx. 10% faster with `compact_str` for that crate, presumably a result of increasing the allocation threshold from 23 to 24 bytes.

### DIFF
--- a/compact_str/src/lib.rs
+++ b/compact_str/src/lib.rs
@@ -26,6 +26,7 @@ use core::str::{
     FromStr,
     Utf8Error,
 };
+use std::borrow::Cow;
 
 mod asserts;
 mod features;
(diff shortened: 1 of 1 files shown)


ParkMyCar:
Thanks for the PR! 

I'd be curious to hear how the migration from `smartstring` to `compact_str` was. I know there are still some APIs missing from `CompactStr` that are present in `String`.

mcronce:
For my use cases, at least, it's just been a matter of (along with adding this `From` impl) `s/smartstring::alias::String/compact_str::CompactStr/g`

PR #71, opened 2022-03-27
refactor: Rename CompactStr to CompactString

ParkMyCar:
This PR renames `CompactStr` to `CompactString`

Some feedback I've received is using the "str" suffix isn't great because it suggests that a `CompactStr` doesn't own the underlying string, just like a `&str`. We change the suffix to "String" to fix this and improve the understanding.

We also add a type alias of `CompactStr` so existing crates will continue to work, albeit with a deprecation warning. This breaking change will go out with `v0.4` to prevent suddenly introducing warnings in projects that depend on `compact_str`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 # Upcoming
+* Rename `CompactStr` to `CompactString`, add deprecation note for `CompactStr`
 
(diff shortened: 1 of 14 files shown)


ParkMyCar:
Closed in favor of #97

PR #73, opened 2022-03-31
Implement fmt::Write for CompactStr

Kijewski:
This makes it possible to `write!()` into a CompactStr. This is useful if you know that the formatted string will be (most often) shorter than 24 bytes.

### DIFF
--- a/compact_str/src/lib.rs
+++ b/compact_str/src/lib.rs
@@ -678,4 +678,11 @@ impl Extend<String> for CompactStr {
     }
 }
 
+impl fmt::Write for CompactStr {
+    fn write_str(&mut self, s: &str) -> fmt::Result {
+        self.push_str(s);
+        Ok(())
+    }
+}
+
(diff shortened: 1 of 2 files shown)


ParkMyCar:
Thanks for the PR! It's probably fine to bump the MSRV, let me just look into this a bit more

ParkMyCar:
Ahh I see, the optimization provided by `Arguments::as_str()` is just when the arguments contain 0 or 1 static strings and no arguments to format into the resulting string.

Given it's an optimization in only the most trivial cases, I'd opt to keep the MSRV at 1.49, and if you could move the commented code into a draft PR like you suggested, that would be great :)

Kijewski:
Yeah, I only added the `.as_str()` case for "completeness". It should occur seldom enough, and more likely in generated code than in what a person writes.

I removed the comment, and will open the draft PR. :)

ParkMyCar:
Thanks for making the changes, and for the PR!

PR #76, opened 2022-04-01
Optimize fmt::Write for trivial cases

Kijewski:
This PR adds an optimization to CompactStr's fmt::Write implementation for invocations like that extend the buffer by a single string, e.g. `write!(s, "example")`

### DIFF
--- a/compact_str/src/lib.rs
+++ b/compact_str/src/lib.rs
@@ -683,6 +683,16 @@ impl fmt::Write for CompactStr {
         self.push_str(s);
         Ok(())
     }
+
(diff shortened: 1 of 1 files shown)


Kijewski:
Arguments::as_str() only became stable in 1.52. If the MSRV gets increased, this PR can be picked, too.

NobodyXu:
@Kijewski Maybe you can use [rustversion](https://github.com/dtolnay/rustversion) to automatically apply the optimization for rust >= 1.52?

ParkMyCar:
Implemented in #107

PR #77, opened 2022-04-01
Tell miri to tag raw pointers

CAD97:
Stricter provenance, yo

Basically, tells miri to pay more attention to stacked borrows w.r.t. raw pointers. (Also forbids int2ptr as a side effect, but if we can avoid it, that's a [good thing](https://github.com/rust-lang/rust/issues/95228).)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,6 +10,7 @@ name: CI
 env:
   RUSTFLAGS: "-D warnings"
   PROPTEST_CASES: 10000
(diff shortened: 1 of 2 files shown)


ParkMyCar:
Sweeeeet!

I haven't had a chance to look too much into Miri other than to know that it exists. Glancing at other flags I also see `miri-check-number-validity`, it seems like that could also be beneficial to enable?

CAD97:
Yeah, check number validity is also a good one to enable if your code is compatible with it. I just didn't in this PR because local runs were stalling and I wasn't sure why.

PR #79, opened 2022-04-01
Stricten miri, again

CAD97:
Check number validity, i.e. forbid transmute(ptr, int).


ParkMyCar:
Thanks! I always appreciate stricter tests!

PR #81, opened 2022-04-05
impl a bunch of Add<T>s for CompactStr, and Add<CompactStr> for String

mcronce:


compact_str/src/lib.rs
```diff
@@ -685,4 +688,52 @@ impl fmt::Write for CompactStr {
...
+}
+
+impl Add<CompactStr> for String {
+    type Output = Self;
+    fn add(mut self, rhs: CompactStr) -> Self::Output {
+        self.push_str(&rhs);
+        self
+    }
+}
+
```
ParkMyCar, line 738: could we also `impl Add<&CompactStr> for String`?

ParkMyCar, line 738: could we also `impl Add<&CompactStr> for String`?

ParkMyCar:
Thanks for the PR!

PR #83, opened 2022-04-26
tests: Ignore failing test on PowerPC 64-bit

ParkMyCar:
There's an issue with formatting floats on PowerPC Little Endian 64-bit, for now we'll ignore the relevant assertions for this platform to keep the X-Plat CI target high signal

### DIFF
--- a/compact_str/src/tests.rs
+++ b/compact_str/src/tests.rs
@@ -390,8 +390,6 @@ macro_rules! assert_int_MAX_to_compact_str {
 }
 
 #[test]
-// TODO: Fix bug in powerpc64, which is a little endian system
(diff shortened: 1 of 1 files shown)


compact_str/src/tests.rs
```diff
@@ -390,6 +390,8 @@ macro_rules! assert_int_MAX_to_compact_str {
 }
 
 #[test]
+// TODO: Fix bug in powerpc64, which is a little endian system
```
NobodyXu, line 393: The tracking issue for this is https://github.com/rust-lang/rust/issues/96306

NobodyXu (changes requested):

PR #84, opened 2022-04-26
Feature/speedup github actions

NobodyXu:
Speedup github actions using `actions/cache@v2` and `cargo-nextest` for testing.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,6 +69,15 @@ jobs:
           toolchain: nightly
           override: true
           components: miri
+      - uses: actions/cache@v2
+        with:
+          path: |
+            ~/.cargo/bin/
+            ~/.cargo/registry/index/
+            ~/.cargo/registry/cache/
+            ~/.cargo/git/db/
+            target/
+          key: ${{ github.event.repository.name }}-${{ runner.os }}-cargo-miri-v2
       - name: Run Miri
         run: |
           cargo miri test --all-features --manifest-path=compact_str/Cargo.toml
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,10 +47,14 @@ jobs:
             ~/.cargo/git/db/
             target/
           key: ${{ github.event.repository.name }}-${{ runner.os }}-cargo-test-v2
-      - uses: actions-rs/cargo@v1
+      - uses: actions-rs/install@v0.1
+        name: Install cargo-nextest
         with:
-          command: test
-          args: --release --all-features --manifest-path=compact_str/Cargo.toml -- --include-ignored
+          crate: cargo-nextest
+          version: latest
+      - name: cargo test
+        run: |
+          cargo nextest run --release --all-features --manifest-path=compact_str/Cargo.toml -- --include-ignored
 
   test-nightly:
     name: cargo test nightly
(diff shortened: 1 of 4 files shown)


ParkMyCar:
@NobodyXu it looks like CI broke because your changes to `ci.yml` and `cross_platform.yml` are invalid and preventing the files from being parsed

ParkMyCar:
Thanks for your work here! I'll take a closer look soon 🙂

NobodyXu:
> Thanks for your work here! I'll take a closer look soon 🙂

This PR basically adds caching to the jobs and uses [`cargo-nextest`] for testing.

[`cargo-nextest`] [executes all tests in parallel](https://nexte.st/book/how-it-works.html)

[`cargo-nextest`]: https://nexte.st/

.github/workflows/ci.yml
```diff
@@ -38,25 +38,47 @@ jobs:
...
         with:
-          command: test
-          args: --release --all-features --manifest-path=compact_str/Cargo.toml -- --include-ignored
+          path: |
+            ~/.cargo/bin/
+            ~/.cargo/registry/index/
+            ~/.cargo/registry/cache/
+            ~/.cargo/git/db/
+            target/
+          key: ${{ github.event.repository.name }}-${{ runner.os }}-cargo-test-v3
```
ParkMyCar, line 49: I don't think we need the repo name here, but we should include a hash of the `Cargo.lock`, [example](https://github.com/actions/cache/blob/main/examples.md#rust---cargo). This way if the dependencies change it invalidates the cache

.github/workflows/ci.yml
```diff
@@ -38,25 +38,47 @@ jobs:
...
+      - name: Install cargo-nextest
+         run: cargo install cargo-nextest
+      - name: cargo test
+        run: |
+          cargo nextest run --release --all-features --manifest-path=compact_str/Cargo.toml -- --include-ignored
```
ParkMyCar, line 54: I'd prefer to stick with normal `cargo test` for now as `nextest` can actually cause slowdowns on Windows and macOS. It's a very promising test running suite though!

.github/workflows/ci.yml
```diff
@@ -38,25 +38,47 @@ jobs:
           profile: minimal
           toolchain: stable
           override: true
-      - uses: actions-rs/cargo@v1
+      - uses: actions/cache@v2
```
ParkMyCar, line 41: any reason we shouldn't use `cache@v3`?

NobodyXu: No, it's simply because I didn't know `cache@v3` is available, I will update that.

.github/workflows/ci.yml
NobodyXu: How about using `cargo-nextest` for linux, but continue to use `cargo-test` for windows and macOS?

.github/workflows/ci.yml
NobodyXu: Done!

.github/workflows/ci.yml
NobodyXu: Done!

.github/workflows/ci.yml
NobodyXu: OK, I've reverted to using `cargo-test` for windows and macOS.

Though TBF, I've been using `cargo-nextest` for macOS on M1 and it is much faster to run tests for this crate and `tokio-rs/bytes`.

NobodyXu:
> @NobodyXu it looks like CI broke because your changes to `ci.yml` and `cross_platform.yml` are invalid and preventing the files from being parsed

There is indeed a syntax error regarding indentation in `ci.yml` and `cross_platform.yml`.

```
- name: ...
   run:
```

yml rejects the code above and requires me to write:

```
- name: ...
  run:
```

NobodyXu:
@ParkMyCar This PR is now ready for merge!

Looking at the times taken, this PR indeed speedup the github actions.

For `cargo-miri`, fuzzing and x-plat, using cache can speedup the workflow to roughly 7-8 minutes, while originally takes roughly 9 minutes or longer.

For job `cargo-test` and `cargo-test-nightly`, the speedup is even more significant thanks to caching plus the use of `cargo-nextest`.

It takes roughly 30s for these jobs.

One thing to note that is the time listed above is all performed on the exact same codebase with no modification.

Once deployed, the actual improvement might be smaller than the number listed above, but never the less, I think this PR would make checks in PR run faster.

ParkMyCar:
Thanks for doing this work! Merging!

PR #85, opened 2022-04-30
tests: Add fuzzing and unit tests for ToCompactStr

ParkMyCar:
Like the title says, adds fuzz coverage and more unit tests for testing the `ToCompactStr` trait


compact_str/src/tests.rs
```diff
@@ -370,6 +370,141 @@ fn test_plus_operator() {
...
+#[should_panic]
+fn test_f64_scientific() {
+    let val: f64 = 1.3594035585e-314;
+
+    let compact = val.to_compact_str();
+    let word = val.to_string();
+
+    assert_eq!(compact, word);
+}
+
```
NobodyXu, line 507: Just my two cents:

Maybe we should also have `i128` and `u128` tests here?

compact_str/src/tests.rs
```diff
@@ -370,6 +370,141 @@ fn test_plus_operator() {
     assert_eq!(String::from("a") + CompactStr::from("b"), "ab");
 }
```
NobodyXu, line 372: So we basically unit test the integer and float specialisation of `ToCompactStr` here.

Note that we also have specialisation for `NonZero*` newtype wrappers of the integer, `char`, `String` and `CompactStr`.

fuzz/src/lib.rs
```diff
@@ -34,6 +37,44 @@ pub enum Creation<'a> {
...
     FromBoxStr(Box<str>),
+    /// Create from a number using [`ToCompactStr`]
+    NumToCompactStr(NumType),
+    /// Create from a boolean using [`ToCompactStr`]
+    BoolToCompactStr(bool),
+    /// Create from a char using [`ToCompactStr`]
+    CharToCompactStr(char),
+    /// Create from a String using [`ToCompactStr`]
+    StringToComapctStr(String),
+}
```
NobodyXu, line 48: Note that we also have specialisation for `CompactStr` in `ToCompactStr`.

fuzz/src/lib.rs
```diff
@@ -180,6 +221,64 @@ impl Creation<'_> {
...
+                    NumType::I128(val) => (val.to_compact_str(), val.to_string()),
+                    // NumType::F32(val) => (val.to_compact_str(), val.to_string()),
+                    // NumType::F64(val) => (val.to_compact_str(), val.to_string()),
+                };
+
+                assert_eq!(compact, word);
+
+                // Numbers should always be inlined, except 64 or 128 bit integers, they might be
+                // heap allocated
+                match num_type {
```
NobodyXu, line 245: I think it is possible to do this:

```suggestion
                match num_type {
                    NumType::U128(_) | NumType::I128(_) => {
                        assert_properly_allocated(&compact, &word);
                    }
                    #[cfg(target_pointer_width = "32")]
                    NumType::U64(_) | NumType::I64(_) => {
                        assert_properly_allocated(&compact, &word);
                    }
                    _ => assert!(!compact.is_heap_allocated()),
                };
```

NobodyXu (changes requested):

NobodyXu:
There are really a ton of changes of binary files for the fuzzing.

compact_str/src/tests.rs
ParkMyCar: Good idea, added!

compact_str/src/tests.rs
ParkMyCar: Good call out for the `NonZero*` types, I added those to the fuzzing

fuzz/src/lib.rs
ParkMyCar: Changed this to a different general case

fuzz/src/lib.rs
ParkMyCar: Good callout!

ParkMyCar:
> There are really a ton of changes of binary files for the fuzzing.

Yeah those are the corpus/seed files, there are other strategies for them, e.g. uploading them somewhere like S3, but for now it's just easy to keep them in the repo

NobodyXu:
> > There are really a ton of changes of binary files for the fuzzing.
> 
> Yeah those are the corpus/seed files, there are other strategies for them, e.g. uploading them somewhere like S3, but for now it's just easy to keep them in the repo

If they are auto generated, is it possible for it to be generated in the github action then cached?

PR #86, opened 2022-04-30
tests: Remove cargo-nextest because it wasn't running anything

ParkMyCar:
I just realized the workflows that use `cargo-nextest` are completing so fast because it's not actually running any tests, [example](https://github.com/ParkMyCar/compact_str/runs/6240801199?check_suite_focus=true). For now I'm just removing nextest because it's not obvious why it's not actually testing anything

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,12 +47,11 @@ jobs:
             ~/.cargo/git/db/
             target/
           key: ${{ runner.os }}-${{ hashFiles('**/Cargo.lock') }}-cargo-test-v3
-      - name: Install cargo-nextest
-        run: |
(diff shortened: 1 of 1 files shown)


sunshowers:
Hi there, author of nextest here. nextest wasn't running any tests because it was filtering tests on the substring "--include-ignored". Instead, you need to pass in --run-ignored=all (without the -- since it's an argument to nextest, not the test runner).

ParkMyCar:
ah ha! Thanks for the solution @sunshowers 🙂, opened #87

sunshowers:
@ParkMyCar Inanna says hi!

ParkMyCar:
@sunshowers tell Inanna I say hi back! I need to chat with her soon about some randomized testing stuff

PR #91, opened 2022-05-05
Reenable float tests on powerpc64le for `ToCompactStr`

NobodyXu:
Last time I tried to reproduce the bug on nightly, it is already fixed.

### DIFF
--- a/.github/workflows/msrv.yml
+++ b/.github/workflows/msrv.yml
@@ -45,4 +45,4 @@ jobs:
       - name: cargo test msrv..
         run: |
           cd compact_str
-          cargo hack test --feature-powerset --optional-deps --version-range 1.49..
+          cargo hack check --feature-powerset --optional-deps --version-range 1.49..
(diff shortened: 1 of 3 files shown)


NobodyXu:
Turns out that `ryu` still cannot properly format `3.2_f32` but can format `f64` correctly https://github.com/dtolnay/ryu/issues/47

Also, the tests cannot be compiled on `1.49` due to `structmeta` requires `2021` edition.

NobodyXu:
I also discovered some new bugs on `rustc` nightly, which formats `f32::NEG_INFINITY` as `inf` instead of `-inf`.

NobodyXu:
> Also, the tests cannot be compiled on 1.49 due to structmeta requires 2021 edition.

I fixed this by running `cargo-check` instead of `cargo-test` in workflow `msrv1`.

NobodyXu:
This PR has been very out of date and the regression still happens from time to time, so I am closing it for now.

PR #93, opened 2022-05-16
Fix afl jobs in workflow `fuzz`: Do not cache `~/.cargo/bin`

NobodyXu:
Just use the `afl` compiled using the nightly toolchain installed.


ParkMyCar:
Sweet, thanks!

PR #95, opened 2022-05-17
perf: Reduce copies in ToCompactStr for integer types

ParkMyCar:
This PR improves the performance of the `trait ToCompactStr` by reducing the number of copies necessary when converting an integer.

Locally I see at least 15% - 46% improvements


compact_str/src/traits.rs
```diff
@@ -0,0 +1,108 @@
...
+/// if the `Display` implementation returns an error.
+/// This indicates an incorrect `Display` implementation
+/// since `std::fmt::Write for String` never returns an error itself and
+/// the implementation of `ToCompactStr::to_compact_str` only panic if the `Display`
+/// implementation is incorrect..
+impl<T: fmt::Display> ToCompactStr for T {
+    #[inline]
+    fn to_compact_str(&self) -> CompactStr {
+        let repr = match_type!(self, {
+            u8 as s => s.into_repr(),
```
NobodyXu, line 64: `&self` cannot be `u8`, it can only be a reference.

compact_str/src/traits.rs
```diff
@@ -0,0 +1,108 @@
...
+            &u8 as s => s.into_repr(),
+            i8 as s => s.into_repr(),
```
NobodyXu, line 66: Same for all the other types here, it cannot be take by value, but by reference only.

NobodyXu (changes requested):

NobodyXu:
@ParkMyCar According to my understanding this PR optimize `ToCompactStr` specialisation for integers, by directly format `u8`, `i8`, `u16`, `i16`, `u32` and `i32` into the `CompactStr` inline storage directly.

For `u64` and `i64`, they are formatted into `CompactStr` inline storage on 64-bit platform.

For `u128` and `i128`, they are formatted using `std::format!` and might require heap allocation.
But doesn't that mean formatting of 128-bit integer could be much slower than other integer types?

compact_str/src/traits.rs
NobodyXu: Note that it could be `&&u8` or `&&&u8`, but not sure it is worth the effort to optimize for these cases.

ParkMyCar: Ah that makes sense, thanks for the help here!

ParkMyCar: Ah that makes sense, thanks for the help here!

ParkMyCar:
> @ParkMyCar According to my understanding this PR optimize `ToCompactStr` specialisation for integers, by directly format `u8`, `i8`, `u16`, `i16`, `u32` and `i32` into the `CompactStr` inline storage directly.
> 
> For `u64` and `i64`, they are formatted into `CompactStr` inline storage on 64-bit platform.
> 
> For `u128` and `i128`, they are formatted using `std::format!` and might require heap allocation. But doesn't that mean formatting of 128-bit integer could be much slower than other integer types?

I updated the implementation, 8, 16, and 32 bit integers should always be inlined, 64 bit integers will always be inlined on 64-bit machines, and possibly on 32-bit machines, depending on how many digits the number has.

Good call with the 128-bit integers. I re-added the dependency on `itoa` and use that for the impl of `ToCompactStr`. It turns out determining the str length of a 128 bit int takes longer than the memcpy, so we use your original impl of writing to a `itoa::Buffer` and then copying

ParkMyCar:
> @ParkMyCar According to my understanding this PR optimize `ToCompactStr` specialisation for integers, by directly format `u8`, `i8`, `u16`, `i16`, `u32` and `i32` into the `CompactStr` inline storage directly.
> 
> For `u64` and `i64`, they are formatted into `CompactStr` inline storage on 64-bit platform.
> 
> For `u128` and `i128`, they are formatted using `std::format!` and might require heap allocation. But doesn't that mean formatting of 128-bit integer could be much slower than other integer types?

I updated the implementation, 8, 16, and 32 bit integers should always be inlined, 64 bit integers will always be inlined on 64-bit machines, and possibly on 32-bit machines, depending on how many digits the number has.

Good call with the 128-bit integers. I re-added the dependency on `itoa` and use that for the impl of `ToCompactStr`. It turns out determining the str length of a 128 bit int takes longer than the memcpy, so we use your original impl of writing to a `itoa::Buffer` and then copying

PR #100, opened 2022-05-28
Fix provenance issues

Nilstrieb:
When creating a `BoxString` using `from_string` or `from_box_str`, `str::as_ptr` was used to get the pointer, which only has read provenance for the initialized part of the string. Going through `Vec` in the `from_string` and through `Box::into_raw` in the `from_box_str` case fixed the issue.

Using proper miri flags (`MIRIFLAGS=-Zmiri-strict-provenance`) miri also complains about `repr::arc`, which is wrong since it uses a zero length array to be a fake custom dst, which is not allowed.


Nilstrieb:
The MSRV check fails because of `addr_of_mut`, which was added in 1.51.
There are 3 options to fix this:
- Raise the MSRV
- Create a reference (unsound)
- Do manual pointer arithmetic since it's repr(C)

.github/workflows/ci.yml
```diff
@@ -10,7 +10,7 @@ name: CI
 env:
   RUSTFLAGS: "-D warnings"
   PROPTEST_CASES: 10000
-  MIRIFAGS: "-Zmiri-tag-raw-pointers -Zmiri-check-number-validity"
+  MIRIFLAGS: "-Zmiri-strict-provenance"
```
ParkMyCar, line 13: can we re-add `-Zmiri-check-number-validity`

compact_str/src/repr/arc/mod.rs
```diff
@@ -6,12 +6,15 @@ use std::sync::atomic::Ordering;
 use std::{
```
ParkMyCar, line 6: this module (`arc`) isn't actually used any more, all of the code is dead. I kept it around to possibly use in the future, but it really should get deleted.

I'll make a change to delete the module, or you can go ahead and delete it too. This should remove the need for the `addr_of_mut` API, which should fix the MSRV issue!

compact_str/src/repr/boxed/mod.rs
```diff
@@ -1,4 +1,5 @@
 use core::iter::Extend;
```
ParkMyCar, line 1: Looks great!

ParkMyCar (changes requested):
Thanks for making these changes! Just see the one comment around deleting the `arc` module since it's not used anymore, after that we should be good to go!

compact_str/src/repr/arc/mod.rs
Nilstrieb: I'll delete it.

ParkMyCar: Thanks!

ParkMyCar (approved):
Thanks for the PR!

PR #101, opened 2022-06-07
Add macro_rules `format_compact!`

NobodyXu:
So that users can format any args into `CompactString` similar to how they can do it using [`std::format!`](https://doc.rust-lang.org/std/macro.format.html).

### DIFF
--- a/compact_str/src/macros.rs
+++ b/compact_str/src/macros.rs
@@ -37,5 +37,8 @@ mod tests {
     fn test_macros() {
         assert_eq!(format_compact!("2"), "2");
         assert_eq!(format_compact!("{}", 2), "2");
+
+        assert!(!format_compact!("2").is_heap_allocated());
+        assert!(!format_compact!("{}", 2).is_heap_allocated());
     }
 }
--- a/compact_str/src/macros.rs
+++ b/compact_str/src/macros.rs
@@ -14,8 +14,8 @@
 /// depending on the intended destination of the string.
 ///
 /// To convert a single value to a string, use the
-/// [`ToCompactString::to_compact_string`] method, which uses
-/// the [`Display`] formatting trait.
+/// `ToCompactString::to_compact_string` method, which uses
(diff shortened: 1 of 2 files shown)


compact_str/src/macros.rs
```diff
@@ -0,0 +1,41 @@
+/// Creates a String using interpolation of runtime expressions.
```
ParkMyCar, line 1: Replace String with `CompactString`

compact_str/src/macros.rs
```diff
@@ -0,0 +1,41 @@
...
+macro_rules! format_compact {
+    ($($arg:tt)*) => {{
+        $crate::ToCompactString::to_compact_string(&format_args!($($arg)*))
+    }}
+}
+
+#[cfg(test)]
+mod tests {
+    #[test]
+    fn test() {
```
ParkMyCar, line 37: rename to `test_macro()` or something more descriptive

compact_str/src/macros.rs
```diff
@@ -0,0 +1,41 @@
...
+        assert_eq!(format_compact!("2"), "2");
+        assert_eq!(format_compact!("{}", 2), "2");
```
ParkMyCar, line 39: assert the values returned from `format_compact!` are inlined, since in this case they're short enough where they should be

ParkMyCar (changes requested):

compact_str/src/macros.rs
NobodyXu: Done!

compact_str/src/macros.rs
NobodyXu: Done!

compact_str/src/macros.rs
NobodyXu: Done!

ParkMyCar (approved):

ParkMyCar:
Thanks @NobodyXu for the PR!

PR #75, opened 2022-03-31
Smol option

CAD97:
#22 but up to date with git main. (Missing some extra tests from #22, though.)

Attempts to avoid the perf cliff seen by #22 by dealing almost exclusively in "`ReprNiche`", and only using `ReprUnion` as a view into `ReprNiche`.

( 👏 on figuring out how to use UTF-8 to get a full 24 bytes inline, btw! )

PR #105, opened 2022-06-17
perf: Option<CompactString> same size as CompactString

ParkMyCar:
Changes `CompactString` such that `Option<CompactString>` is the same size! 

Specifically `size_of::<CompactString>() == size_of::<Option<CompactString>>() == size_of::<String>()`. We're able to achieve this by defining an enum `NonMaxU8` where the valid values are `[0,255)`, this allows the compiler to use 255 to store the `None` variant of the `Option<_>`.

Combination of #22 and #75 

@CAD97 I pulled in your changes from #75 and then updated some comments and fuzzing tests, what do you think?


compact_str/src/repr/discriminant.rs
```diff
@@ -21,7 +19,15 @@ pub struct DiscriminantMask {
 impl DiscriminantMask {
     #[inline(always)]
     pub const fn discriminant(&self) -> Discriminant {
-        if self.val == HEAP_MASK {
+        if self.val == u8::MAX {
+            // HACK: This allows us to make assertions within a `const fn` without requiring
+            // nightly, see unstable `const_panic` feature. This results in a build
+            // failure, not a runtime panic
```
CAD97, line 25: `const_panic` was stabilized in 1.57; it's probably worth it to require 1.57 since we're bumping the MSRV anyway.

Also, to be clear: this can still be a runtime panic if the function is called at runtime (not in a const context).

compact_str/src/repr/mod.rs
```diff
@@ -13,31 +13,58 @@ mod boxed;
...
-pub union Repr {
+/// Used as a discriminant to identify different variants
+pub const HEAP_MASK: u8 = 0b11111110;
+
+/// This is the "compiler facing" representation for the struct that underpins `CompactString`. The
+/// odd layout enables the compiler to represent an `Option<CompactString>` in the same amount of
+/// bytes as `CompactString`. In other words, it allows the compiler to see a "niche" value, which
+/// the compiler then uses to store the `None` variant.
+///
+/// The `size_of::<Repr>()` needs to be the same as `size_of::<String>()`, we construct that with
```
CAD97, line 42:
```suggestion
/// The `size_of::<Repr>()` needs to be the same as `size_of::<CompactString>()`, we construct that with
```

The *needs* is to be the same size as `CompactString`. We additionally *want* the size to be the same as `String`, but this is a design choice, not a requirement.

compact_str/src/repr/mod.rs
```diff
@@ -243,21 +269,45 @@ impl Repr {
...
                 // SAFETY: We checked the discriminant to make sure the union is `inline`
-                MutStrongRepr::Inline(unsafe { &mut self.inline })
+                MutStrongRepr::Inline(unsafe { &mut self.as_union_mut().inline })
             }
         }
     }
+
+    #[inline(always)]
+    const fn from_inline(repr: InlineString) -> Self {
+        // SAFETY: An `InlineString` and `Repr` have the same size and alignment
```
CAD97, line 283: Nit: alignment doesn't matter for by-value `tranamute`.

CAD97 (approved):
I can't comment on the fuzzing harness, but otherwise everything looks reasonable. Just a couple minor suggestions/nits

compact_str/src/repr/discriminant.rs
ParkMyCar: This makes sense to me! Updated the MSRV to 1.57 and now just call `panic!(...)`

compact_str/src/repr/mod.rs
ParkMyCar: I referenced `String` here instead of `CompactString`, since a `CompactString` is just a wrapper around `Repr`, so the size of a `Repr` should always equal `CompactString`. Re-worded this a bit to hopefully describe our intents better

compact_str/src/repr/mod.rs
ParkMyCar: Updated!

ParkMyCar:
@CAD97 thanks for the review!

I've been fuzzing these changes for the past 4 hours with 3 different machines, each with a different CPU arch (M1 Max, Ubuntu + Intel i7, Raspbian + Pi 4). I'm going to keep running the fuzzer, but I'm pretty confident in the changes, thanks for the help here!

ParkMyCar:
cc @NobodyXu, now that we're bumping the MSRV we can merge #76, was there anything else we wanted to do once the MSRV was increased?

NobodyXu:
> cc @NobodyXu, now that we're bumping the MSRV we can merge #76, was there anything else we wanted to do once the MSRV was increased?

No, I don't have anythin on my mind.

PR #118, opened 2022-06-19
feat: add impl From<CompactString> for String

### LINKED ISSUE #117: impl From<CompactString> for String
As the title says, we should allow users to easily and efficiently convert from a CompactString into a String

neoeinstein:
Tests added and run under `miri`. Ensures that no reallocation/copy of already-boxed strings happens in a round-trip, with the address, length, and capacity remaining stable.

Closes #117

### DIFF
--- a/compact_str/src/repr/boxed/mod.rs
+++ b/compact_str/src/repr/boxed/mod.rs
@@ -162,10 +162,7 @@ impl BoxString {
         // * `length` is less than or equal to capacity, due to internal invaraints.
         // * `capacity` is correctly maintained internally.
         // * `BoxString` only ever contains valid UTF-8.
-        // * `BoxString` capacity is always on the heap.
-        unsafe {
-            String::from_raw_parts(this.ptr.as_ptr(), this.len, this.cap.as_usize_unchecked())
-        }
+        unsafe { String::from_raw_parts(this.ptr.as_ptr(), this.len, this.capacity()) }
     }
 
     #[inline]
--- a/compact_str/src/repr/boxed/mod.rs
+++ b/compact_str/src/repr/boxed/mod.rs
@@ -612,8 +612,10 @@ mod tests {
     fn test_32_bit_max_inline_cap() {
         // 65 is the ASCII value of 'A'
         // `SIXTEEN_MB - 2` is the max value we can store for capacity inline, when on 32-bit archs
-        let word_buf = vec![65; SIXTEEN_MB - 2];
-        let string = String::from_utf8(word_buf).unwrap();
+        let word_buf = vec![b'A'; SIXTEEN_MB - 2];
+        // SAFETY: We've just created the vector above, and it contains only ASCII characters.
+        // This speeds up `miri` runs significantly.
+        let string = unsafe { String::from_utf8_unchecked(word_buf) };
 
         let box_string = BoxString::new(&string);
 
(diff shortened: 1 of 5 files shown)


compact_str/src/repr/boxed/mod.rs
```diff
@@ -151,6 +151,20 @@ impl BoxString {
...
+        // Ensure that we don't free the data we are transferring to `String`
+        let this = ManuallyDrop::new(self);
+
+        // SAFETY:
+        // * The memory in `ptr` was previously allocated by the same allocator the standard library
+        //   uses, with a required alignment of exactly 1.
+        // * `length` is less than or equal to capacity, due to internal invaraints.
+        // * `capacity` is correctly maintained internally.
+        // * `BoxString` only ever contains valid UTF-8.
+        unsafe { String::from_raw_parts(this.ptr.as_ptr(), this.len, this.capacity()) }
```
ParkMyCar, line 165: A `BoxedString` can store the capacity on either the stack _or_ the heap, we store the capacity on the heap when there isn't enough room on the stack, see [this doc](https://github.com/ParkMyCar/compact_str/blob/main/compact_str/src/repr/boxed/capacity.rs#L18-L36) comment for an explanation.

tl;dr I think you'll have to change this to something like:
```
match self.cap.as_usize() {
  // capacity is on the stack, form a string from these parts
  Ok(capacity) => {
    let this = ManuallyDrop::new(self);
    unsafe { String::from_raw_parts(this.ptr.as_ptr(), this.len, capacity) }
  },
  // capacity is on the heap! we need to form a new String
  Err(_) => {
    // ... drop the existing string and form a new one
  }
}
```

compact_str/src/repr/mod.rs
```diff
@@ -501,6 +537,12 @@ impl<'a> MutStrongRepr<'a> {
     }
 }
 
+#[derive(Debug)]
+enum StrongIntoRepr {
```
ParkMyCar, line 541: nit: maybe `StrongOwnedRepr`? makes it a little more general in case we use this for more than just `into_string()` in the future

compact_str/src/tests.rs
```diff
@@ -695,3 +705,131 @@ fn test_to_compact_string() {
         format_compact!("0{}67890123456789{}", "12345", 999999)
     );
 }
+
+#[test]
```
ParkMyCar, line 709: love the existing test coverage, thank you!

Could you add one more test for the 16MB edge case that exists for 32-bit machines? see [here](https://github.com/ParkMyCar/compact_str/blob/main/compact_str/src/repr/boxed/mod.rs#L608) for an example

ParkMyCar (changes requested):
Thanks for the PR, looking real good! Just some small changes necessary to handle an edge case with 32-bit arch's

compact_str/src/repr/boxed/mod.rs
```diff
@@ -593,8 +612,10 @@ mod tests {
     fn test_32_bit_max_inline_cap() {
         // 65 is the ASCII value of 'A'
         // `SIXTEEN_MB - 2` is the max value we can store for capacity inline, when on 32-bit archs
-        let word_buf = vec![65; SIXTEEN_MB - 2];
-        let string = String::from_utf8(word_buf).unwrap();
+        let word_buf = vec![b'A'; SIXTEEN_MB - 2];
+        // SAFETY: We've just created the vector above, and it contains only ASCII characters.
+        // This speeds up `miri` runs significantly.
+        let string = unsafe { String::from_utf8_unchecked(word_buf) };
```
ParkMyCar, line 618: Fantastic!

ParkMyCar (approved):
Great job! Thanks for the PR and speeding up some of our tests!

compact_str/src/repr/boxed/mod.rs
ParkMyCar: Good idea to re-use the existing buffer. I think this is something we can iterate on, for now I filed issue #120 to track this

CAD97: > the layouts aren't compatible

Yeah, you'll run into an issue as `String`'s buffer is `u8` aligned but the inline length is `usize` aligned. *At a performance cost*, we could store the length unaligned to make the allocated alignment 1.

The unstable `Allocator` trait allows *requesting* (though an implementation can theoretically refuse to fulfill) a change in alignment of an allocation. The stable `GlobalAlloc` does not.

My vote is to just copy/realloc, and put a note tl if/when `Allocator` is stabilized optimize it to opportunistically use a layout-adjusting realloc so the underlying allocator can avoid the copy.

PR #124, opened 2022-06-24
Add `as_mut_str()` method

Kijewski:
`String::as_mut_str()` returns a mutable slice of its contents. This PR adds the same method to CompactString.

The method is safe, because no safe operation on `&mut str` can result in a broken UTF-8 string.


ParkMyCar (approved):
This looks great, thank you!

PR #125, opened 2022-06-24
Implement `replace_range()`

Kijewski:
This PR implements `CompactString::replace_range()` to work the same way as the method in String of the same name.

It uses DerefMut as implemented in #124, but it could easily be written to work without it if #124 is not accepted.

### DIFF
--- a/compact_str/src/lib.rs
+++ b/compact_str/src/lib.rs
@@ -26,8 +26,10 @@ use core::hash::{
 use core::iter::FromIterator;
 use core::ops::{
     Add,
+    Bound,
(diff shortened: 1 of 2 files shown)


ParkMyCar (approved):
This looks great, would you mind rebasing off of `main`, and then we can merge? It looks like this PR also includes the changes from #124.

Side note: thanks for updating the fuzzing harness here and in #124!

Kijewski:
Rebased. Thank you for the library. :) It's always so much easier to contribute when a project already uses testing and fuzzing.

PR #126, opened 2022-06-25
fuzz: Add more coverage for various APIs

ParkMyCar:
This PR adds more fuzzing coverage of various APIs. I used LLVM's code coverage tools to determine what APIs aren't currently covered by fuzzing, and started fuzzing them!

### DIFF
--- a/fuzz/src/creation.rs
+++ b/fuzz/src/creation.rs
@@ -1,0 +1,529 @@
+//! Different ways in which we can create a [`CompactString`] and a "control" [`String`]. If we can
+//! successfully generate this pair, we then run varios actions on them, which are defined in the
+//! [`super::actions`] module.
+
+use std::borrow::Cow;
+use std::collections::VecDeque;
+use std::io::Cursor;
+use std::num;
+use std::str::FromStr;
+
+use arbitrary::Arbitrary;
+use compact_str::{
+    CompactString,
+    CompactStringExt,
+    ToCompactString,
+};
+
+use super::assert_properly_allocated;
+use crate::MAX_INLINE_LENGTH;
+
+#[derive(Arbitrary, Debug)]
+pub enum Creation<'a> {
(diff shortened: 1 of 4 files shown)


fuzz/src/creation.rs
```diff
@@ -0,0 +1,529 @@
...
+                    _ => panic!("CompactString and core::str read UTF-8 differently?"),
+                }
+            }
+            BufUnchecked(data) => {
+                let mut buffer = Cursor::new(data);
+
+                // The data provided might not be valid UTF-8. We mainly want to make sure we don't
+                // panic, and the data is written correctly. Before returning either of these types
+                // we'll make sure they contain valid data
+                let compact = unsafe { CompactString::from_utf8_buf_unchecked(&mut buffer) };
```
Kijewski, line 302: Is this right? An exactly 24 bytes random string would be broken the result if `data[23] >= 192`, wouldn't it?

ParkMyCar: Great observation! We [check](https://github.com/ParkMyCar/compact_str/blob/main/compact_str/src/repr/bytes.rs#L44-L55) for this exact case when reading the bytes from a `Buf`, and if we hit it, we expand to a heap string which fixes the issue.

Kijewski: Ah, nice! I did not assume that you're checking the `unchecked()` data. :) But I guess it's the right way to do that, because you are allowed to build illegal Strings, and otherwise CompactString would not be a "proper" replacement.

ParkMyCar: Valid assumption to make :) 

When I first encountered this problem I did think twice about adding the check, since the API is literally called "unchecked". But I figured it was the right thing to do because the check is very cheap, comparing a single byte on the stack, and the bug it fixes is very dangerous, possibly de-referencing a garbage pointer

PR #130, opened 2022-06-27
fix/feat: Change the existing Add<...> impls

ParkMyCar:
Implementing `Add<CompactString> for String` breaks `String + &String`, as long as `compact_str` is used somewhere in your current crate. Specifically this code won't compile:

```
use compact_str;

fn main() {
  println!("{}", "Hello".to_string() + &(" World".to_string()));
}
```

This isn't good! This has to do with Rust's current trait resolver and is a known issue https://github.com/rust-lang/rust/issues/77143

To fix this, we remove `impl Add<CompactString> for String`. You can still add `CompactString` to `String`, you just need to take a reference to it, then the `impl Add<&str> for String` kicks in.

We also remove a lot of `impl Add<...> for CompactString`, leaving just `impl Add<&str> for CompactString`, which allows us to add _more_ string types than we currently can, e.g. `&Box<str>` and `&Cow<str>`. This is because the compiler auto derefs these types to `&str` and uses that impl

### DIFF
--- a/compact_str/src/lib.rs
+++ b/compact_str/src/lib.rs
@@ -879,50 +879,10 @@ impl fmt::Write for CompactString {
     }
 }
 
-impl Add<Self> for CompactString {
-    type Output = Self;
-    fn add(mut self, rhs: Self) -> Self::Output {
-        self.push_str(&rhs);
-        self
-    }
-}
-
-impl Add<&Self> for CompactString {
-    type Output = Self;
-    fn add(mut self, rhs: &Self) -> Self::Output {
-        self.push_str(rhs);
-        self
-    }
-}
-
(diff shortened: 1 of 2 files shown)


ParkMyCar:
Note: these `Add<...>` impls were first available on `v0.4.0`. This fix has been back ported to:

 * `v0.4.0` -> `v0.4.1`
 * `v0.5.0` -> `v0.5.1`

Technically this was a breaking change because you can no longer add a `String` to a `CompactString`, but to fix it all you need to do is take a reference, e.g. make it `&String`. IMO this small change is worth it to prevent the bug, where importing `compact_str` can break other `String +` operations, even if they don't use `CompactString` at all themselves.

PR #132, opened 2022-06-28
Implement `truncate()`

### LINKED ISSUE #131: feature: truncate
Great work on this library! So far substituting `String` with `CompactString` has given immediate performance improvements in the projects I've tested it on. 

The only unfortunate thing I ran into was that `CompactString` has no `truncate` method like that of `String`. Implementing it user-side is a bit scary because some `unsafe` is likely involved. Is there any chance it could be added in a future update?

Simple example using the `String` api:
```rs
    let mut s = String::from("Hello World!\n\n");
    assert!(s.len() == 14);
    
    s.truncate(s.trim_end().len());
    assert!(s.len() == 12);
```

Kijewski:
Closes #131

### DIFF
--- a/compact_str/src/lib.rs
+++ b/compact_str/src/lib.rs
@@ -634,6 +634,39 @@ impl CompactString {
             }
         }
     }
+
+    /// Truncate the [`CompactString`] to a shorter length.
+    ///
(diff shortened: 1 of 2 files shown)


ParkMyCar (approved):
Wooo! Looks great, thank you!

PR #147, opened 2022-06-30
feat: Implement the Arbitrary trait from various crate

ParkMyCar:
This PR implements the `Arbitrary` trait from the [`arbitrary`](https://github.com/rust-fuzz/arbitrary), [`proptest`](https://github.com/AltSysrq/proptest), and [`quickcheck`](https://github.com/BurntSushi/quickcheck) crates. It then makes those crates optional dependencies.

This makes it easier for folks to include a `CompactString` in their types, and then use randomized testing to cover those types

### DIFF
--- a/compact_str/src/features/arbitrary.rs
+++ b/compact_str/src/features/arbitrary.rs
@@ -1,0 +1,52 @@
+//! Implements the [`arbitrary::Arbitrary`] trait for [`CompactString`]
+
+use arbitrary::{
+    Arbitrary,
+    Result,
+    Unstructured,
+};
+
+use crate::CompactString;
+
+impl<'a> Arbitrary<'a> for CompactString {
+    fn arbitrary(u: &mut Unstructured<'a>) -> Result<Self> {
+        <&str as Arbitrary>::arbitrary(u).map(CompactString::new)
+    }
+
+    fn arbitrary_take_rest(u: Unstructured<'a>) -> Result<Self> {
+        <&str as Arbitrary>::arbitrary_take_rest(u).map(CompactString::new)
+    }
+
+    #[inline]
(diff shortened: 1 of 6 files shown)


compact_str/src/features/arbitrary.rs
```diff
@@ -0,0 +1,43 @@
...
+    Arbitrary,
+    Result,
+    Unstructured,
+};
+
+use crate::CompactString;
+
+impl<'a> Arbitrary<'a> for CompactString {
+    fn arbitrary(u: &mut Unstructured<'a>) -> Result<Self> {
+        <&str as Arbitrary>::arbitrary(u).map(|s| CompactString::new(s))
```
NobodyXu, line 13:
```suggestion
        <&str as Arbitrary>::arbitrary(u).map(CompactString::new)
```

compact_str/src/features/arbitrary.rs
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    fn arbitrary_take_rest(u: Unstructured<'a>) -> Result<Self> {
+        <&str as Arbitrary>::arbitrary_take_rest(u).map(|s| CompactString::new(s))
```
NobodyXu, line 17:
```suggestion
        <&str as Arbitrary>::arbitrary_take_rest(u).map(CompactString::new)
```

NobodyXu (changes requested):

compact_str/src/features/arbitrary.rs
ParkMyCar: Good call, thanks!

ParkMyCar:
Thanks for the review @NobodyXu!

PR #149, opened 2022-07-02
Implement `clear()`

### LINKED ISSUE #133: feat: Implement the clear() API
This issue for tracking the implementation of the [`clear()`](https://doc.rust-lang.org/std/string/struct.String.html#method.clear) method that exists on `String`.

We should implement it exactly as it exists on `String`, that is, it removes all the contents of the `CompactString` setting the length to 0, but does not touch the capacity.

Kijewski:
Closes #133.


ParkMyCar (approved):
Sweet! Thank you for the PR!

PR #153, opened 2022-07-04
Implement `drain()`

### LINKED ISSUE #134: feat: Implement the drain(range) API
This issue for tracking the implementation of the [`drain(range)`](https://doc.rust-lang.org/std/string/struct.String.html#method.drain) method that exists on `String`.

We should implement it exactly as it exists on `String`, that is, it removes the specified range from the `CompactString`, returning all the removed characters as an iterator.

Kijewski:
Makes `replace_range()` reusable, and closes #134.

### DIFF
--- a/compact_str/src/lib.rs
+++ b/compact_str/src/lib.rs
@@ -515,6 +515,45 @@ impl CompactString {
         self.repr.is_heap_allocated()
     }
 
+    /// Ensure that the given range is inside the set data, and that no codepoints are split.
+    ///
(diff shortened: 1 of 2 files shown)


ParkMyCar:
@Kijewski thanks for all of the PRs! I'm away on holiday with family, will review all of these in a few days :)

Kijewski:
No rush, have a nice holiday! :)

ParkMyCar (approved):
Wooo! Great work with this!

PR #155, opened 2022-07-06
Implement `shrink_to()` and `shrink_to_fit()`

### LINKED ISSUE #143: feat: Implement the shrink_to(...) API
This issue for tracking the implementation of the [`shrink_to(...)`](https://doc.rust-lang.org/std/string/struct.String.html#method.shrink_to) method that exists on `String`.

### LINKED ISSUE #144: feat: Implement the shrink_to_fit(...) API
This issue for tracking the implementation of the [`shrink_to_fit(...)`](https://doc.rust-lang.org/std/string/struct.String.html#method.shrink_to_fit) method that exists on `String`.

Kijewski:
Closes #143, closes #144.

### DIFF
--- a/compact_str/src/repr/boxed/mod.rs
+++ b/compact_str/src/repr/boxed/mod.rs
@@ -400,7 +400,7 @@ impl BoxString {
     ///
     /// Returns the new capacity as `Ok(…)` if the string was reallocated.
     /// Or the old capacity as `Err(…)` if the string could not be reallocated.
-    fn realloc(&mut self, new_capacity: usize) -> Result<usize, ()> {
+    pub fn realloc(&mut self, new_capacity: usize) -> Result<usize, ()> {
         // can't reallocate a BoxString to a size less than the current length
         if new_capacity < self.len {
             return Err(());
--- a/compact_str/src/tests.rs
+++ b/compact_str/src/tests.rs
@@ -17,6 +17,8 @@ const MAX_SIZE: usize = 24;
 #[cfg(target_pointer_width = "32")]
 const MAX_SIZE: usize = 12;
 
+const SIXTEEN_MB: usize = 16 * 1024 * 1024;
+
 /// generates random unicode strings, upto 80 chars long
 pub fn rand_unicode() -> impl Strategy<Value = String> {
     proptest::collection::vec(proptest::char::any(), 0..80).prop_map(|v| v.into_iter().collect())
@@ -850,7 +852,6 @@ fn test_into_string_large_string_with_excess_capacity() {
 
 #[test]
 fn test_into_string_where_32_bit_capacity_is_on_heap() {
-    const SIXTEEN_MB: usize = 16 * 1024 * 1024;
     let buf = vec![b'a'; SIXTEEN_MB - 1];
     // SAFETY: `buf` is filled with ASCII `a`s.
     // This primarily speeds up miri, as we don't need to check every byte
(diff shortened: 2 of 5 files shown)


Kijewski:
This PR implements `BoxString::realloc()` instead a custom of reallocation. Depending on the used `#[global_allocator]` this might nor might not be an improvement, but it should never be worse. The method could be used by reserve(), too.

compact_str/src/repr/boxed/mod.rs
```diff
@@ -407,6 +410,32 @@ impl BoxString {
             }
         }
     }
+
+    /// Tries to reallocate the buffer, and sets the new capacity on success.
+    ///
+    /// Returns the new capacity as `Ok(…)` if the string was reallocated.
+    /// Or the old capacity as `Err(…)` if the string could not be reallocated.
+    pub unsafe fn realloc(&mut self, new_capacity: usize) -> Result<usize, usize> {
+        let old_capacity = self.capacity();
```
ParkMyCar, line 419: if you call `self.cap.as_usize()` that will return an error when the capacity is on the heap. For now it would be totally fine with me to just `return Err(...)` in that case to avoid having to handle the change in layouts

ParkMyCar (changes requested):
Thanks for the PR @Kijewski, sorry it took me so long to review!

compact_str/src/tests.rs
```diff
@@ -1017,3 +1017,33 @@ fn test_insert() {
         "\u{ffff}\u{ffff}\u{ffff}\u{ffff}\u{ffff}\u{ffff}\u{ffff}\u{ffff}",
     );
 }
+
+#[test]
+fn test_shrink_to_fit() {
```
ParkMyCar, line 1022: could you add a test that creates a string of length 16MB + 1, and then shrink it to say 1MB? And then another test where we create a string of say 20MB and then shrink it to a size of 20MB - 1?

I want to make sure we cover the case two cases:
1. shrinking from a "capacity on heap" -> "capacity inlined"
2. shrinking from a "capacity on heap" -> "capacity on heap"

Kijewski:
Oh, thanks for the review! I missed the layout change when the capacity hits 16MB. Maybe it would be best to delay this PR, and implement a realloc() method first? One that takes care of the corner cases, so it can be used for growing, too?

Afaict there's are 3 by 3 grid of transitions {inline, capacity on stack, capacity on heap}^2. From the same to the same layout; and from and to inline are the simple transitions.

| from\to | inline | cap on stack | cap on heap |
|---|---|---|---|
| inline | no-op | malloc | malloc |
| cap on stack | free | reallocate | free + malloc |
| cap on heap | free | free + malloc | reallocate |

ParkMyCar:
> Oh, thanks for the review! I missed the layout change when the capacity hits 16MB. Maybe it would be best to delay this PR, and implement a realloc() method first? One that takes care of the corner cases, so it can be used for growing, too?
> 
> Afaict there's are 3 by 3 grid of transitions {inline, capacity on stack, capacity on heap}^2. From the same to the same layout; and from and to inline are the simple transitions.
> from\to 	inline 	cap on stack 	cap on heap
> inline 	no-op 	malloc 	malloc
> cap on stack 	free 	reallocate 	free + malloc
> cap on heap 	free 	free + malloc 	reallocate

This sounds like a great plan! I've started a work-in-progress implementation in #160

ParkMyCar:
Hey @Kijewski, now that #160 has been merged do you want to take another look at this? No worries if not :)

Kijewski:
I swear I meant to take a look at this PR. :D Updated. I am not 100% sure that `test_reserve_shrink_roundtrip()` is a very useful test, but I guess it should find errors for 32bit platforms.

ParkMyCar:
Woooo! Looks great, thanks for the PR @Kijewski :)

PR #156, opened 2022-07-06
Deduplicate `CompactString::push()`

Kijewski:
It's simply encoding the character as a UTF-8 sequence, then appending the sequence. So `push()` can simply call `push_str()`.


ParkMyCar:
Sweet! Thanks for the code cleanup :)

PR #160, opened 2022-07-17
perf: Implement realloc for BoxString

ParkMyCar:
This PR implements `realloc(size)` for `BoxString`, which is the underlying heap representation of a `CompactString`

### DIFF
--- a/compact_str/src/repr/boxed/mod.rs
+++ b/compact_str/src/repr/boxed/mod.rs
@@ -6,6 +6,8 @@ use core::{
     slice,
     str,
 };
+use std::alloc;
+use std::ptr::NonNull;
 
 mod capacity;
 use capacity::Capacity;
@@ -13,6 +15,17 @@ use capacity::Capacity;
 mod inner;
 
 const MIN_SIZE: usize = core::mem::size_of::<usize>() / 2;
+
+/// [`BoxString`] grows at an amortized rates of 1.5x
+///
+/// Note: this is different than [`std::string::String`], which grows at a rate of 2x. It's debated
+/// which is better, for now we'll stick with a rate of 1.5x
+#[inline(always)]
+fn amortized_growth(cur_len: usize, additional: usize) -> usize {
+    let required = cur_len + additional;
+    let amortized = 3 * cur_len / 2;
+    core::cmp::max(amortized, required)
+}
 
 #[repr(C)]
 pub struct BoxString {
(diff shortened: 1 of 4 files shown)


compact_str/src/repr/boxed/mod.rs
```diff
@@ -14,6 +16,17 @@ mod inner;
...
 
+/// [`BoxString`] grows at an amortized rates of 1.5x
+///
+/// Note: this is different than [`std::string::String`], which grows at a rate of 2x. It's debated
+/// which is better, for now we'll stick with a rate of 1.5x
+#[inline(always)]
+fn amortized_growth(cur_len: usize, additional: usize) -> usize {
+    let required = cur_len + additional;
+    let amortized = 3 * cur_len / 2;
+    core::cmp::max(amortized, required)
```
Kijewski, line 27: Should this function maybe be made fallible on overflows? I.e. using checked_add?

compact_str/src/repr/boxed/mod.rs
```diff
@@ -694,28 +757,100 @@ mod tests {
...
-        #[cfg(target_pointer_width = "32")]
-        assert!(box_string.cap.as_usize().is_err());
-        // on 64-bit archs it's still inlined
-        #[cfg(not(target_pointer_width = "32"))]
-        assert!(box_string.cap.as_usize().is_ok());
+        cfg_if::cfg_if! {
+            if #[cfg(target_pointer_width = "64")] {
+                // on 64-bit archs it's still inlined
+                assert!(box_string.cap.as_usize().is_ok());
+            } else {
```
Kijewski, line 777: I think the `target_pointer_width = "32"` check should stay, and `compile_error!()` if the pointer width is neither. Some seem to work on porting rust to [msp430-none-elf](https://doc.rust-lang.org/nightly/rustc/platform-support.html), which is a 16bit platform.

compact_str/src/repr/boxed/mod.rs
ParkMyCar: I'm not totally sure what we should do here yet, I left a TODO to think about the design of overflows

compact_str/src/repr/boxed/mod.rs
ParkMyCar: Good idea! I updated all the tests to use `compile_error!()`

ParkMyCar:
Thanks for the review @Kijewski!

PR #191, opened 2022-07-25
feat: Implement CompactString::remove()

### LINKED ISSUE #141: feat: Implement the remove(...) API
This issue for tracking the implementation of the [`remove(...)`](https://doc.rust-lang.org/std/string/struct.String.html#method.remove) method that exists on `String`.

ParkMyCar:
* add unit tests
* add doc tests
* add prop tests
* add to fuzzing harness

Fixes #141

### DIFF
--- a/compact_str/src/lib.rs
+++ b/compact_str/src/lib.rs
@@ -507,6 +507,72 @@ impl CompactString {
         self.repr.push_str(s)
     }
 
+    /// Removes a [`char`] from this [`CompactString`] at a byte position and returns it.
+    ///
+    /// This is an *O*(*n*) operation, as it requires copying every element in the
+    /// buffer.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `idx` is larger than or equal to the [`CompactString`]'s length,
+    /// or if it does not lie on a [`char`] boundary.
+    ///
+    /// # Examples
+    ///
+    /// ### Basic usage:
+    ///
+    /// ```
+    /// # use compact_str::CompactString;
+    /// let mut c = CompactString::from("hello world");
+    ///
+    /// assert_eq!(c.remove(0), 'h');
+    /// assert_eq!(c, "ello world");
+    ///
+    /// assert_eq!(c.remove(5), 'w');
+    /// assert_eq!(c, "ello orld");
+    /// ```
+    ///
+    /// ### Past total length:
+    ///
+    /// ```should_panic
+    /// # use compact_str::CompactString;
+    /// let mut c = CompactString::from("hello there!");
+    /// c.remove(100);
+    /// ```
+    ///
+    /// ### Not on char boundary:
+    ///
+    /// ```should_panic
+    /// # use compact_str::CompactString;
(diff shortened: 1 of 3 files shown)


compact_str/src/lib.rs
```diff
@@ -507,6 +507,67 @@ impl CompactString {
...
+    ///
+    /// ```should_panic
+    /// # use compact_str::CompactString;
+    /// let mut c = CompactString::from("🦄");
+    /// c.remove(1);
+    /// ```
+    #[inline]
+    pub fn remove(&mut self, idx: usize) -> char {
+        // get the char
+        let ch = match self[idx..].chars().next() {
```
Kijewski, line 553: You could use [`char_indices()`](https://doc.rust-lang.org/stable/std/primitive.str.html#method.char_indices) instead of  `chars()` to omit the `len_utf8()` call.

Kijewski: You could use `let s = self.as_mut_str();` and operate on `s`, so the pointers don't have to be calculated multiple times, and miri understands what you are doing.

Kijewski: I meant something like that:

```rust
pub fn remove(&mut self, idx: usize) -> char {
    let s = self.as_mut_str();
    let old_len = s.len();

    let substr = &mut s[idx..];
    let (ch_len, ch) = substr.char_indices()
        .next()
        .expect("cannot remove a char from the end of a string");

    // shift everything back one character
    unsafe {
        core::ptr::copy(
            substr.as_ptr().add(ch_len),
            substr.as_mut_ptr(),
            old_len - (idx + ch_len),
        );
        self.set_len(old_len - ch_len);
    }
    ch
}
```

ParkMyCar: Ahh I see, thanks for providing the example :)

Here `ch_len` isn't the length of the char returned, I believe it would actually always be 0 because it's the first character of `substr`. 

We run Miri with `-Zmiri-strict-provenance` and getting two pointers at once seems to violate the stacked borrow rules. Although I like your idea of operating on `substr`, I refactored the impl to work similar to what you've described.

Kijewski: Oops, you're absolutely right. I confused myself how `char_indices()` works. :)

fuzz/src/actions.rs
```diff
@@ -204,6 +206,19 @@ impl Action<'_> {
                 assert_eq!(control.as_str(), compact.as_str());
                 assert_eq!(compact.capacity(), compact_capacity);
             }
+            Remove(val) => {
+                let idx = to_index(control, val);
```
Kijewski, line 210: `to_index()` can also return `s.len()`. This should have caused some crashes in the fuzzer.

ParkMyCar: Good call out! It looks like as it exists today `to_index()` calls `.cycle()` before it `.chain([s.len()])`, which was causing `s.len()` to never get emitted, since `.cycle()` never ends. I updated `to_index()` to fix this and added tests to make sure it's doing what we want

PR #199, opened 2022-08-08
`cargo clippy --fix -- -D clippy::pedantic`

Kijewski:


Kijewski:
I cannot restart the failed test. It had a "spurious network error", so the error should be unrelated to the PR.

PR #207, opened 2022-08-14
Implement `from_utf16le`, `from_utf16be`, `from_utf16ne`

Kijewski:


compact_str/src/lib.rs
```diff
@@ -1170,12 +1173,98 @@ impl CompactString {
...
+    /// If the slice has an odd number of bytes, or if it did not contain valid UTF-16 characters,
+    /// a [`Utf16Error`] is returned.
+    ///
+    /// # Examples
+    ///
+    /// Please see [`CompactString::from_utf16le()`] and [`CompactString::from_utf16be()`]
+    /// for examples.
+    #[inline(always)]
+    #[cfg(any(target_endian = "little", target_endian = "big"))]
+    pub fn from_utf16ne(v: &[u8]) -> Result<Self, Utf16Error> {
```
ParkMyCar, line 1262: Thinking about this a bit further, do you think we actually need a native endian version of this function?

I haven't worked with UTF-16 that much, but my impression is if someone needs to be specific about the endianness that's because some protocol specifies UTF-16LE or UTF-16BE. I want to prevent any overlap between `from_utf16(buf: &[u16])` and `from_utf16ne(buf: &[u8])` to prevent confusion about which method should be used

compact_str/src/lib.rs
```diff
@@ -1170,12 +1173,98 @@ impl CompactString {
...
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use compact_str::CompactString;
+    /// const DANCING_MEN: &[u8] = b"\x3d\xd8\x6f\xdc\x0d\x20\x42\x26\x0f\xfe";
+    /// let dancing_men = CompactString::from_utf16le(DANCING_MEN).unwrap();
+    /// assert_eq!(dancing_men, "👯‍♂️");
+    /// ```
+    pub fn from_utf16le(v: &[u8]) -> Result<Self, Utf16Error> {
```
ParkMyCar, line 1226: nit: can this be generic over some type that implements `AsRef<[u8]>`, e.g. `from_utf16le<B: AsRef<[u8]>>(buf: B)` to be consistent with the other `from_*(...)` APIs

compact_str/src/lib.rs
```diff
@@ -1170,12 +1173,98 @@ impl CompactString {
...
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use compact_str::CompactString;
+    /// const DANCING_WOMEN: &[u8] = b"\xd8\x3d\xdc\x6f\x20\x0d\x26\x40\xfe\x0f";
+    /// let dancing_women = CompactString::from_utf16be(DANCING_WOMEN).unwrap();
+    /// assert_eq!(dancing_women, "👯‍♀️");
+    /// ```
+    pub fn from_utf16be(v: &[u8]) -> Result<Self, Utf16Error> {
```
ParkMyCar, line 1245: Same as above, can this be generic over some type that implements `AsRef<[u8]>`

compact_str/src/lib.rs
```diff
@@ -1170,12 +1173,98 @@ impl CompactString {
...
         result
     }
+
+    #[inline(always)]
+    fn from_utf16x(
+        v: &[u8],
+        from_x: impl Fn(u16) -> u16,
+        from_x_bytes: impl Fn([u8; 2]) -> u16,
+    ) -> Result<Self, Utf16Error> {
+        match unsafe { v.align_to::<u16>() } {
```
ParkMyCar, line 1189: nit: Document safety invariants

compact_str/src/lib.rs
```diff
@@ -1170,12 +1173,98 @@ impl CompactString {
...
+            (&[], _, &[_]) | (&[_], _, &[]) => {
+                // Input had an odd number of bytes.
+                Err(Utf16Error(()))
+            }
+            (&[], v, _) => {
```
ParkMyCar, line 1194: The [stdlib](https://doc.rust-lang.org/std/primitive.slice.html#method.align_to) states:
> The method may make the middle slice the greatest length possible for a given type and input slice, but only your algorithm’s performance should depend on that, not its correctness. It is permissible for all of the input data to be returned as the prefix or suffix slice.

So it seems while in the common case `v` should contain all of the data from our original slice, it's not guaranteed.

Could we instead do something like:

```
// Input had an odd number of bytes
if v.len() % 2 != 0 {
  return Err(...)
}

match unsafe { v.align_to::<u16>() } {
  // input is correctly aligned
  (&[], v, &[]) => {
    core::char::decode_utf16(v.iter().copied().map(from_x)
      .collect::<Result<CompactString, _>>()
      .map_err(|_| Utf16Error(()))
  },
  // alignment is off
  _ => {
    // SAFETY: we can always reinterpret a `[u8; 2*N]` slice as `[[u8; 2]; N]`
    let v = unsafe { slice::from_raw_parts(v.as_ptr().cast(), v.len() / 2) };
    std::char::decode_utf16(v.iter().copied().map(from_x_bytes))
        .collect::<Result<CompactString, _>>()
        .map_err(|_| Utf16Error(()))
  }
}
```

ParkMyCar (changes requested):
Thanks for the PR! The test setup looks great! Just a few comments :)

compact_str/src/lib.rs
Kijewski: Made generic.

compact_str/src/lib.rs
Kijewski: Made generic.

compact_str/src/lib.rs
Kijewski: Done.

compact_str/src/lib.rs
Kijewski: Ah, I didn't see that. Fixed.

Kijewski:
Thank you for your review! :) I made the requested changes. I also saw the comment in `from_utf16()` about <https://github.com/rust-lang/rust/issues/48994>, so I don't use `collect::<Result<_, _>>()`.

Yes, with `from_utf16ne()` we would give the user a footgun to create platform incompatible code. The input should contain a BOM, or the user should use some other means to tell the endianness.

PR #208, opened 2022-08-17
Add rkyv serialization

Kijewski:


Kijewski:
I omitted writing fuzzing tests. We would only test if rkyv can (de)serialize `&str`s, which would be of little use, IMO. What do you think?

compact_str/src/features/rkyv.rs
```diff
@@ -0,0 +1,85 @@
...
+
+impl PartialOrd<CompactString> for ArchivedString {
+    #[inline]
+    fn partial_cmp(&self, other: &CompactString) -> Option<std::cmp::Ordering> {
+        PartialOrd::partial_cmp(self.as_str(), other.as_str())
+    }
+}
+
+#[cfg(test)]
+#[cfg_attr(miri, ignore)] // https://github.com/rust-lang/unsafe-code-guidelines/issues/134
```
ParkMyCar, line 63: Does this stacked borrows violation come from our code, or rkyv's?

ParkMyCar (approved):

ParkMyCar:
> I omitted writing fuzzing tests. We would only test if rkyv can (de)serialize `&str`s, which would be of little use, IMO. What do you think?

Not adding to the fuzzing framework is fine with me, we don't have fuzz coverage for `serde` either. Would you mind adding a `proptest` though?

ParkMyCar:
Thanks for the PR! I haven't used `rkyv` myself but have definitely heard of it, and making `CompactString` usable with more crates is always good :)

compact_str/src/features/rkyv.rs
Kijewski: rkyv. <https://github.com/rust-lang/unsafe-code-guidelines/issues/134#issuecomment-1087628968>

Actually, I think I see what the problem is in their code. In [repr.rs:40-62](https://github.com/rkyv/rkyv/blob/f552fabc204ebc27f812166e636a75d45d97cba1/rkyv/src/string/repr.rs#L40-L62) they would need to replace `self.inline.bytes.as_ptr()` with `self.inline.bytes[..].as_ptr()` to make MIRI understand what is going on. If I don't forget it, I'll try if that fixes the problem.

ParkMyCar:
> > I omitted writing fuzzing tests. We would only test if rkyv can (de)serialize `&str`s, which would be of little use, IMO. What do you think?
> 
> Not adding to the fuzzing framework is fine with me, we don't have fuzz coverage for `serde` either. Would you mind adding a `proptest` though? You could also convert the existing test if that works

@Kijewski sorry for the quick succession of comments here. Just realized we don't have __any__ tests for `serde`! I'm going to add some tests now, and will add a proptest for `rkyv` at the same time

PR #210, opened 2022-08-18
Implement `from_utf16ne_lossy` and `from_utf16be_lossy`

### LINKED ISSUE #136: feat: Implement the from_utf16_lossy(...) API
This issue for tracking the implementation of the [`from_utf16_lossy(...)`](https://doc.rust-lang.org/std/string/struct.String.html#method.from_utf16_lossy) method that exists on `String`.

Kijewski:
Closes #136.

### DIFF
--- a/compact_str/src/lib.rs
+++ b/compact_str/src/lib.rs
@@ -1191,10 +1191,10 @@ impl CompactString {
         }
 
         // Note: we don't use collect::<Result<_, _>>() because that fails to pre-allocate a buffer,
-        // even though the size of our iterator, `buf`, is known ahead of time.
+        // even though the size of our iterator, `v`, is known ahead of time.
         //
         // rustlang issue #48994 is tracking the fix
-        let mut result = CompactString::with_capacity(v.len());
+        let mut result = CompactString::with_capacity(v.len() / 2);
 
         // SAFETY: `u8` and `u16` are `Copy`, so if the alignment fits, we can transmute a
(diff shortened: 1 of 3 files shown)


compact_str/src/lib.rs
```diff
@@ -1191,10 +1191,10 @@ impl CompactString {
         }
 
         // Note: we don't use collect::<Result<_, _>>() because that fails to pre-allocate a buffer,
-        // even though the size of our iterator, `buf`, is known ahead of time.
+        // even though the size of our iterator, `v`, is known ahead of time.
         //
         // rustlang issue #48994 is tracking the fix
-        let mut result = CompactString::with_capacity(v.len());
+        let mut result = CompactString::with_capacity(v.len() / 2);
```
Kijewski, line 1197: I forgot the `/ 2` in my last PR.

* `v.len() / 2` assumes that most of the input is ASCII.
* `v.len()` assumes that most of the input non-western European scripts in the [BMP](https://codepoints.net/basic_multilingual_plane), e.g. Korean, Japanese, Chinese, Russian …
* `v.len() / 2 * 3` assumes that most of input is text in uncommon or dead scripts
* `v.len() * 2` assumes that most of the input is emojis 💩

ParkMyCar:
Looks great! Thanks for the PR!

compact_str/src/lib.rs
ParkMyCar: Good call out, thanks for the update!

PR #217, opened 2022-08-22
Enable "proptest/std" feature

Kijewski:
The last build on docs.rs failed:
<https://docs.rs/crate/compact_str/0.6.0/builds/615366>.
To use proptest, you have to either enable "std" or "alloc". Because
this project uses features in "std", we can enable "proptest/std". We
know that the support is there.

I cannot tell why the bug does not manifest in the CI, which checks the
documentation creation. On my own system it works without this PR, too.


ParkMyCar:
Great catch!

ParkMyCar:
Bumped to `v0.6.1` and the [docs](https://docs.rs/compact_str/0.6.1/compact_str/) look good!

PR #220, opened 2022-08-28
perf: Branchless string accesses

ParkMyCar:
Work in progress, need to work on large strings on 32-bit arch's and code organization. Just putting this up to run against CI

### DIFF
--- a/compact_str/src/repr/intrinsics.rs
+++ b/compact_str/src/repr/intrinsics.rs
@@ -37,8 +37,8 @@ pub fn cmov_ptr_len(
             unsafe {
                 core::arch::asm! {
                     "cmp {d}, 254",
-                    "cmovz {pd}, {ps}",
-                    "cmovz {ld}, {ls}",
+                    "cmovnz {pd}, {ps}",
+                    "cmovnz {ld}, {ls}",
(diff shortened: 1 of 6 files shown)


compact_str/src/repr/intrinsics.rs
```diff
@@ -0,0 +1,108 @@
...
+                    sl = in(reg) length_src,
+                    options(pure, nomem, nostack),
+                };
+            }
+        } else if #[cfg(all(any(target_arch = "x86", target_arch = "x86_64"), not(miri)))] {
+            unsafe {
+                core::arch::asm! {
+                    "cmp {d}, 254",
+                    "cmovnz {pd}, {ps}",
+                    "cmovnz {ld}, {ls}",
```
Kijewski, line 40: cmov looks at the flag register to tell if it should do the mov. So the first mov should always be unconditional, i.e. a normal mov.

Kijewski: Ah, I misinterpreted the code. You are conditionally overwriting a variable, not an register. OK, that's fine :)

ParkMyCar:
Closing while I continue to iterate

PR #229, opened 2022-10-23
perf: Refactor underlying buffers for branchless access

ParkMyCar:
This PR refactors the underlying `InlineString` and `BoxString` structs in a major way to allow for branch-less string accesses. It also renames these structs to `InlineBuffer` and `HeapBuffer`, treating these types strictly as "buffers". We also move string operations into `Repr` (e.g. pushing and popping), which de-dupes code.


compact_str/src/repr2/heap.rs
```diff
@@ -16,6 +16,19 @@ const MIN_HEAP_SIZE: usize = MAX_SIZE + mem::size_of::<usize>();
...
+///
+/// Note: this is different than [`std::string::String`], which grows at a rate of 2x. It's debated
+/// which is better, for now we'll stick with a rate of 1.5x
+///
+/// TODO: Handle overflows in the case of __very__ large Strings
+#[inline(always)]
+pub fn amortized_growth(cur_len: usize, additional: usize) -> usize {
+    let required = cur_len + additional;
+    let amortized = 3 * cur_len / 2;
+    core::cmp::max(amortized, required)
```
NobodyXu, line 29: Nit: Could use `usize::max` instead:

```suggestion
    amortized.max(required)
```

compact_str/src/repr2/heap.rs
```diff
@@ -16,6 +16,19 @@ const MIN_HEAP_SIZE: usize = MAX_SIZE + mem::size_of::<usize>();
...
 
+/// [`HeapBuffer`] grows at an amortized rates of 1.5x
+///
+/// Note: this is different than [`std::string::String`], which grows at a rate of 2x. It's debated
+/// which is better, for now we'll stick with a rate of 1.5x
+///
+/// TODO: Handle overflows in the case of __very__ large Strings
+#[inline(always)]
+pub fn amortized_growth(cur_len: usize, additional: usize) -> usize {
+    let required = cur_len + additional;
```
NobodyXu, line 27: To avoid overflow:

```suggestion
    let required = cur_len.saturating_add(additional);
```

compact_str/src/repr2/heap.rs
```diff
@@ -16,6 +16,19 @@ const MIN_HEAP_SIZE: usize = MAX_SIZE + mem::size_of::<usize>();
...
+    let amortized = 3 * cur_len / 2;
```
NobodyXu, line 28: To avoid overflow:

```suggestion
    let amortized = cur_len.saturating_mul(3) / 2;
```

NobodyXu (changes requested):

compact_str/src/repr2/mod.rs
```diff
@@ -0,0 +1,856 @@
...
+        }
+
+        if needed_capacity <= MAX_SIZE {
+            // It's possible to have a `Repr` that is heap allocated with a capacity less than
+            // MAX_SIZE, if that `Repr` was created From a String or Box<str>
+            //
+            // SAFTEY: Our needed_capacity is >= our length, which is <= than MAX_SIZE
+            let inline = unsafe { InlineBuffer::new(self.as_str()) };
+            // SAFETY: `InlineBuffer` and `Repr` are the same size
+            *self = unsafe { mem::transmute(inline) };
```
NobodyXu, line 296: How about implementing `impl From<InlineBuffer> for Repr` and `impl From<HeapBuffer> for Repr` so that we don't need these `unsafe` transmute blocks everywhere?

compact_str/src/repr2/mod.rs
```diff
@@ -0,0 +1,856 @@
...
+                // * The memory in `ptr` was previously allocated by the same allocator the standard
+                //   library uses, with a required alignment of exactly 1.
+                // * `length` is less than or equal to capacity, due to internal invaraints.
+                // * `capacity` is correctly maintained internally.
+                // * `BoxString` only ever contains valid UTF-8.
+                unsafe { String::from_raw_parts(this.ptr.as_ptr(), this.len, cap) }
+            }
+        } else {
+            let pointer = &self as *const _ as *const u8;
+            let length = core::cmp::min((last_byte.wrapping_sub(LENGTH_MASK)) as usize, MAX_SIZE);
```
NobodyXu, line 223: I've seen this a couple of time, maybe it can be extracted as a separate function?

compact_str/src/repr2/mod.rs
```diff
@@ -0,0 +1,856 @@
...
+/// When our string is stored inline, we represent the length of the string in the last byte, offset
+/// by `LENGTH_MASK`
+pub const LENGTH_MASK: u8 = 0b11000000;
+
+const EMPTY: Repr = Repr::new_inline("");
+
+#[repr(C)]
+pub struct Repr(
+    // We have a pointer in the repesentation to properly carry provenance
+    *const (),
```
NobodyXu, line 38: I think it's possible to use `ptr::NonNull<u8>` here if we modify `push_str`, `extend` and etc to ignore null bytes inserted into the string.

NobodyXu (changes requested):

ParkMyCar:
Thanks for the review @NobodyXu! I'll update the PR soon with the feedback :)

compact_str/src/repr2/mod.rs
ParkMyCar: For now I'm going to stick with this pattern because I believe it'll be more performant. Otherwise we'll need to call through `fn from(...)` which would incur a performance penalty

compact_str/src/repr2/mod.rs
ParkMyCar: Good thought! For now I'll just keep it `*const ()` but we could change this in the future

compact_str/src/repr2/mod.rs
ParkMyCar: Not a bad idea, I'll do this in a follow up PR

PR #230, opened 2022-12-29
impl AsRef<[u8]> for CompactString

mcronce:
Needed this for something earlier tonight and noticed it wasn't implemented - one tiny gap in `CompactString`'s public API compared to std `String`'s


ParkMyCar:
Ahhh sorry we missed this API, thanks for the PR!

I plan to merge https://github.com/ParkMyCar/compact_str/pull/229 soon (among other small fixes), and after that I'll cut another release :)

PR #233, opened 2023-01-03
ci: Disable ARMv7 Workflow

ParkMyCar:
The Raspberry Pi I was using as an ARMv7 runner is currently offline, so for now I'm commenting out any workflows that depend on it

### DIFF
--- a/.github/workflows/cross_platform.yml
+++ b/.github/workflows/cross_platform.yml
@@ -86,23 +86,24 @@ jobs:
           command: miri
           args: test --all-features --manifest-path=compact_str/Cargo.toml
 
-  linux_arm7:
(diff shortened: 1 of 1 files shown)


NobodyXu:
> The Raspberry Pi I was using as an ARMv7 runner is currently offline, so for now I'm commenting out any workflows that depend on it

An alternative could be using `cross`, which works nicely in [`cargo-binstall`](https://github.com/cargo-bins/cargo-binstall).

ParkMyCar:
We already use `cross` for a few things, the goal of the ARMv7 runner was to test on different devices that folks might be using, i.e. a Raspberry Pi :)

PR #235, opened 2023-01-16
Fix a typo in documentation

vbasky:
The documentation should state expects instead of excepts in my opinion.


ParkMyCar:
Great catch, thanks @vbasky!

PR #256, opened 2023-02-16
api: Update From<String> and From<Box<str>> to eagerly inline

### LINKED ISSUE #227: Should From<String> convert to inline if possible?
I was profiling something and noticed that `CompactString::clone` was hitting the allocator consistently even though the strings were small. Turns out that if you convert a String into CompactString using the From/Into trait it'll do so by taking ownership of the backing allocation as expected. But as a side-effect further clones will behave like String::clone _even_ if the strings would fit inline.

If you are using CompactString chances are you expect it to be inlineable most of the time so this comes as a performance pitfall depending on how the CompactString was created.

I suggest changing `From<String>` to convert to inline if possible to avoid this scenario. Doing so may make `From<CompactString> for String` slower but I'll argue that an allocation is expected in this case and any re-use is a bonus. Alternatively the Clone implementation can check if the clone should be inline even if the source is boxed.

One could also argue for full flexibility and that the behavior should be kept and a new function added `CompacString::compact`, but it's both cumbersome to use and most people will never realize it's thing. If such hyperspecific apis are desirable I suggest  a different constructor that keeps the exiting behavior `CompactString::from_string_buffer`.

ParkMyCar:
Currently when converting from a `String` to a `CompactString` using `From<String>` (also for `Box<str>`) we wouldn't inline short strings, this was to avoid needing to de-allocate the original `String`. After some feedback collected in #227 we realized while this was performant, it's not what most users expect to happen.

This PR updates the implementation of `From<String>` and `From<Box<str>>` to inline short strings, and it adds a new API `CompactStr::from_string_buffer`, which retains the current behavior of always re-using the underlying buffer. Note, in #254 we implemented inline-ing during `Clone`, so even if you use `from_string_buffer`, future short strings should get inlined.

Fixes #227

### DIFF
--- a/compact_str/src/repr/mod.rs
+++ b/compact_str/src/repr/mod.rs
@@ -140,12 +140,13 @@ impl Repr {
         repr
     }
 
-    /// Create a [`Repr`] from a [`String`], in `O(1)` time.
+    /// Create a [`Repr`] from a [`String`], in `O(1)` time. We'll attempt to inline the string
+    /// if `should_inline` is `true`
     ///
     /// Note: If the provided [`String`] is >16 MB and we're on a 32-bit arch, we'll copy the
     /// `String`.
     #[inline]
-    pub fn from_string(s: String) -> Self {
+    pub fn from_string(s: String, should_inline: bool) -> Self {
         let og_cap = s.capacity();
         let cap = Capacity::new(og_cap);
 
@@ -165,9 +166,14 @@ impl Repr {
             // We only hit this case if the provided String is > 16MB and we're on a 32-bit arch. We
             // expect it to be unlikely, thus we hint that to the compiler
             capacity_on_heap(s)
-        } else if og_cap == 0 {
+        } else if s.is_empty() {
             // We don't expect converting from an empty String often, so we make this code path cold
             empty()
+        } else if should_inline && s.len() <= MAX_SIZE {
+            // SAFETY: Checked to make sure the length of s will fit inline
+            let inline = unsafe { InlineBuffer::new(s.as_str()) };
+            // SAFETY: `InlineBuffer` and `Repr` are the same size
+            unsafe { mem::transmute(inline) }
         } else {
             let mut s = mem::ManuallyDrop::new(s.into_bytes());
             let len = s.len();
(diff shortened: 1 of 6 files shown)


compact_str/src/repr/mod.rs
```diff
@@ -165,9 +166,14 @@ impl Repr {
             // We only hit this case if the provided String is > 16MB and we're on a 32-bit arch. We
             // expect it to be unlikely, thus we hint that to the compiler
             capacity_on_heap(s)
-        } else if og_cap == 0 {
+        } else if s.is_empty() {
```
NobodyXu, line 169: If `should_inline` is `false` and `s` contains a large pre-allocated buffer, then should we re-use the buffer to avoid alloc?

compact_str/src/repr/mod.rs
```diff
@@ -232,7 +238,7 @@ impl Repr {
     }
 
     #[inline]
-    pub fn from_box_str(s: Box<str>) -> Self {
+    pub fn from_box_str(s: Box<str>, should_inline: bool) -> Self {
```
NobodyXu, line 241: IMO this can be forwarded to `Repr::from_string`

NobodyXu (changes requested):

compact_str/src/repr/mod.rs
ParkMyCar: Yeah good thought, I updated the PR to be based on capacity now instead of length

Reason being, when created a `String` will only allocate as much space as needed. When appending to a string though it grows at a rate of 2x to amortize the cost of re-allocations. This leaves us with 2 scenarios:

1. capacity == length, in which case it doesn't matter which we check
2. capacity > length, this indicates the user appended to the string, or specifically allocated more space, which it seems like we'd want to keep

ParkMyCar: hmmm, actually it seems like `Clone`-ing today actually doesn't preserve capacity, [example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=5a9d2bbfccfcd5f53f027cd3166d5ee0), so my assumption that users would want to preserve capacity might be wrong, let me think more

compact_str/src/repr/mod.rs
ParkMyCar: Good call! I made this change

PR #270, opened 2023-05-18
fix: Don't count bytes in ToCompactString

### LINKED ISSUE #269: `ToCompactString` formats twice
This test:

```rust
#[test]
fn to_compact_string_formats_once() {
    #[derive(Default)]
    struct FormatMeOnce(Cell<bool>);

    impl fmt::Display for FormatMeOnce {
        fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
            if self.0.replace(true) {
                panic!("formatted twice!")
            }

            f.write_str("FormatMeOnce")
        }
    }

    assert_eq!(FormatMeOnce::default().to_compact_string(), "FormatMeOnce");
}
```

fails. This problem occurs in practice with `Itertools::format()`, since formatting consumes the iterator, so trying to format it twice panics.

ParkMyCar:
Previously we had an optimization in `ToCompactString` that we would format a type once to determine how many bytes it would need, create a `CompactString` with that many bytes, and then format the type again. After adding optimizations for primitive types this extra pass isn't needed.

Fixes #269


NobodyXu (approved):

NobodyXu:
msrv failed due to rayon pulled in dev-dep.

IMO msrv should only run cargo-check, not cargo-test

ParkMyCar:
Thanks for the review!

> IMO msrv should only run cargo-check, not cargo-test

Yeah only running cargo-check isn't a bad idea, but I would like to keep running tests as long as possible though to prevent odd edge cases with other crates, e.g. `serde`, `rykv`, etc

NobodyXu (approved):

NobodyXu:
> Yeah only running cargo-check isn't a bad idea, but I would like to keep running tests as long as possible though to prevent odd edge cases with other crates, e.g. `serde`, `rykv`, etc


You need to use nightly cargo option -Z minimum-versions, to use the bare minimum version specified in Cargo.toml

NobodyXu:
@ParkMyCar Please have a look at this https://github.com/NobodyXu/concurrent_arena/blob/main/.github/workflows/minimal.yml

PR #273, opened 2023-07-01
feat: Support O(1) `CompactString::from_static_str`

### LINKED ISSUE #264: Support creating CompactString from &'static str in O(1)
I encounter several scenariors where I want to create a CompactString from &'static str and it would be great if that can be done in O(1) by simply storing a reference to the 'static str

NobodyXu:
Fixed #264

### DIFF
--- a/compact_str/src/repr/mod.rs
+++ b/compact_str/src/repr/mod.rs
@@ -314,6 +314,8 @@ impl Repr {
 
     #[inline]
     pub fn push_str(&mut self, s: &str) {
+        // If `s` is empty, then there's no reason to reserve or push anything
+        // at all.
         if s.is_empty() {
             return;
         }
--- a/compact_str/src/repr/mod.rs
+++ b/compact_str/src/repr/mod.rs
@@ -16,21 +16,25 @@ mod inline;
 mod iter;
 mod last_utf8_char;
 mod num;
+mod static_str;
 mod traits;
 
 use capacity::Capacity;
 use heap::HeapBuffer;
 use inline::InlineBuffer;
 use last_utf8_char::LastUtf8Char;
+use static_str::StaticStr;
 pub use traits::IntoRepr;
 
 /// The max size of a string we can fit inline
 pub const MAX_SIZE: usize = std::mem::size_of::<String>();
 /// Used as a discriminant to identify different variants
 pub const HEAP_MASK: u8 = LastUtf8Char::Heap as u8;
+/// Used for `StaticStr` variant
+pub const STATIC_STR_MASK: u8 = LastUtf8Char::Static as u8;
 /// When our string is stored inline, we represent the length of the string in the last byte, offset
 /// by `LENGTH_MASK`
-pub const LENGTH_MASK: u8 = LastUtf8Char::L0 as u8;
+pub const LENGTH_MASK: u8 = 0b11000000;
 
 const EMPTY: Repr = Repr::new_inline("");
 
@@ -47,6 +51,7 @@ pub struct Repr(
     // ...so that the last byte can be a NonMax, which allows the compiler to see a niche value
     LastUtf8Char,
 );
+static_assertions::assert_eq_size!([u8; MAX_SIZE], Repr);
 
 unsafe impl Send for Repr {}
 unsafe impl Sync for Repr {}
@@ -78,6 +83,13 @@ impl Repr {
         } else {
             panic!("Inline string was too long, max length is `std::mem::size_of::<CompactString>()` bytes");
         }
+    }
+
+    #[inline]
+    pub const fn from_static_str(text: &'static str) -> Self {
+        let repr = StaticStr::new(text);
+        // SAFETY: A `StaticStr` and `Repr` have the same size
+        unsafe { core::mem::transmute(repr) }
     }
 
     /// Create a [`Repr`] with the provided `capacity`
@@ -186,8 +198,6 @@ impl Repr {
     /// Converts a [`Repr`] into a [`String`], in `O(1)` time, if possible
     #[inline]
     pub fn into_string(self) -> String {
-        let last_byte = self.last_byte();
-
         #[cold]
         fn into_string_heap(this: HeapBuffer) -> String {
             // SAFETY: We know pointer is valid for `length` bytes
@@ -198,7 +208,7 @@ impl Repr {
             String::from(s)
         }
 
-        if last_byte == HEAP_MASK {
+        if self.is_heap_allocated() {
             // SAFTEY: we just checked that the discriminant indicates we're a HeapBuffer
             let heap_buffer = unsafe { self.into_heap() };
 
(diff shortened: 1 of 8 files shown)


NobodyXu:
@ParkMyCar Friendly pinging since this PR has been stale for one month.

ParkMyCar:
Hey @NobodyXu thanks for this PR and for your patience! I definitely want to add this feature but haven't had the time to review it yet, I will get to it as soon as I can!

NobodyXu:
@ParkMyCar I've rebased this against main and also updated `README.md`, now it should be ready for review and merge again!

NobodyXu:
MIPS CI failed due to missing component `rust-std`
PowerPC and Linux 32bit CI failed due to link errors.

I don't think they are caused by this PR.

compact_str/src/repr/mod.rs
```diff
@@ -311,6 +314,10 @@ impl Repr {
 
     #[inline]
     pub fn push_str(&mut self, s: &str) {
+        if s.is_empty() {
+            return;
+        }
+
```
ParkMyCar, line 320: Why was this case added? Can you add a comment explaining

compact_str/src/repr/mod.rs
```diff
@@ -346,6 +353,10 @@ impl Repr {
     /// Returns the string content, and only the string content, as a slice of bytes.
     #[inline]
     pub fn as_slice(&self) -> &[u8] {
+        if let Some(s) = self.as_static_str() {
+            return s.as_bytes();
+        }
+
```
ParkMyCar, line 359: This is a performance sensitive method because it's how a user accesses the underlying string, so we want to make it as fast as possible. Adding an if-branch here will negatively effect performance.

Ideally when adding support for `&'static str` there shouldn't be any changes to this method, or to `fn len(...)` below, this should be achievable! Specifically the layout of `StaticStr` and `HeapBuffer` should both have the pointer as their first element and length as their second. Then we can kind of represent the `StaticStr` variant as a "heap variant with 0 capacity". This should allow us to keep branchless accesses.

Let me know what you think! We can merge this as-is and I can follow up with these changes, up to you 😊

compact_str/src/repr/mod.rs
```diff
@@ -381,6 +392,10 @@ impl Repr {
     #[allow(clippy::len_without_is_empty)] // is_empty exists on CompactString
     #[inline]
     pub fn len(&self) -> usize {
+        if let Some(s) = self.as_static_str() {
+            return s.len();
+        }
```
ParkMyCar, line 397: see comment on `fn as_slice(...)`

compact_str/src/repr/static_str.rs
```diff
@@ -0,0 +1,31 @@
...
+
+/// A buffer stored on the stack whose size is equal to the stack size of `String`
+/// The last byte is set to 0.
+#[derive(Copy, Clone)]
+#[repr(C)]
+pub struct StaticStr {
+    pub text: &'static str,
+    #[allow(unused)]
+    discriminant: [u8; DISCRIMINANT_SIZE],
+}
```
ParkMyCar, line 19: What do you think of "deconstructing" the `&'static str` and changing this repr to
```
pub struct StaticStr {
  pub ptr: ptr::NonNull<u8>,
  pub len: usize,
  discriminant: [u8; DISCRIMINANT_SIZE],
}
```

I'll need to double check some of the guarantees around static memory, but I think this layout should work just fine since the static mem should not change for the lifetime of the program.

compact_str/src/tests.rs
```diff
@@ -503,14 +519,14 @@ fn test_extend_packed_from_empty() {
...
-    assert!(compact.is_empty());
-    assert_eq!(compact, "");
+    for mut compact in [CompactString::from(""), CompactString::from_static_str("")] {
+        (0..num_pops).for_each(|_| {
+            let ch = compact.pop();
+            assert!(ch.is_none());
+        });
+        assert!(compact.is_empty());
+        assert_eq!(compact, "");
+    }
```
ParkMyCar, line 529: nit: use the `#[test_case]` macro instead of a for loop here

ParkMyCar (commented):
Wooo! Thanks for this PR @NobodyXu, I'm excited to get this feature merged! Let me know what you think w.r.t. the accessor methods and layout. More than happy to merge as-is and iterate 😊

ParkMyCar:
> MIPS CI failed due to missing component `rust-std` PowerPC and Linux 32bit CI failed due to link errors.
> 
> I don't think they are caused by this PR.

Yeah CI, specifically the steps that use [`cross`](https://github.com/cross-rs/cross), seems busted. I've disabled them for now but will re-enable before the next release.

compact_str/src/repr/mod.rs
NobodyXu: I'm quite busy recently and you are more familiar with compact-str, so I think merging it as-is and then improve it later might be faster and have less mental burden.

compact_str/src/repr/static_str.rs
NobodyXu: Agree, that should work.

compact_str/src/repr/mod.rs
NobodyXu: This is for cases where `s` is empty, there's no need to reserve anything in that case, hence I added this.

compact_str/src/repr/mod.rs
NobodyXu: I've optimized this by changing `StaticStr` layout to be more similiar to `HeapBuffer` so that we can employ the same optimization trick here.

compact_str/src/repr/mod.rs
NobodyXu: Thanks, I've applied the same optimization here.

compact_str/src/repr/static_str.rs
NobodyXu: I've applied this and it indeed works!

compact_str/src/tests.rs
NobodyXu: Thanks, I've applied your suggestion!

compact_str/src/tests.rs
```diff
@@ -536,16 +552,16 @@ fn test_compact_str_is_send_and_sync() {
 fn test_fmt_write() {
     use core::fmt::Write;
 
-    let mut compact = CompactString::default();
-
-    write!(compact, "test").unwrap();
-    assert_eq!(compact, "test");
+    for mut compact in [CompactString::default(), CompactString::from_static_str("")] {
+        write!(compact, "test").unwrap();
+        assert_eq!(compact, "test");
```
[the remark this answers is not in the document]
NobodyXu: Thanks, I've applied your suggestion!

ParkMyCar:
Thanks for the PR @NobodyXu!

NobodyXu:
@ParkMyCar Can we have a new release please?

PR #276, opened 2023-07-17
Open up more niches

Kijewski:
Before this change there is one nieche: The last byte could never have the value `255`. With this change all values in the range `217..=255` cannot occur in the last byte. Actually, these values were already impossible before, but we did not tell the compiler about it.


Kijewski:
I wanted to post my benchmark results, but given that my change made creating `String`s 14% faster, I guess I should put too much stock in criterion's results.

Kijewski:
Should I move the optimization for `Repr::len()` into its own PR?

I thought it was a low-hanging fruit, but I could not get the compiler to unconditionally read `self.1` without `asm!`, which is only stable since Rust 1.57. Even [`ptr::read_volatile()`](https://doc.rust-lang.org/stable/std/ptr/fn.read_volatile.html) gets optimized away. Using [`blackbox()`](https://doc.rust-lang.org/stable/std/hint/fn.black_box.html) works, but is only stable since 1.60, and results in needless writing to the stack.

ParkMyCar:
Thanks @Kijewski this is great!

mcronce:
This is a nice addition - now `Option<cervine::Cow<'_, CompactString, str>>`, `cervine::Cow<'_, CompactString, str>`, `Option<CompactString>`, and `CompactString` are all the same stack size - 24 bytes.

PR #281, opened 2023-07-22
Enable dependabot for cargo and GHA

NobodyXu:


NobodyXu:
The CI failure is an ICE reported here https://github.com/rust-lang/rust/issues/113941 and seems to be already fixed.

ParkMyCar:
Thank you!

PR #283, opened 2023-07-30
Simplify `Repr::len() `

Kijewski:
This PR depends on #276.

Using the fact that `LastUtf8Char::Heap` comes exactly after the last inline length value, we can simplify the calculation a bit.

The resulting code on amd64:

```asm
mov    0x8(%rdi),%rax
movzbl 0x17(%rdi),%ecx
add    $0xffffffffffffff40,%rcx
cmp    $0x18,%rcx
cmovae %rax,%rcx
mov    $0x18,%eax
cmovbe %rcx,%rax
ret
```

### DIFF
--- a/compact_str/src/repr/mod.rs
+++ b/compact_str/src/repr/mod.rs
@@ -381,23 +381,36 @@ impl Repr {
     #[allow(clippy::len_without_is_empty)] // is_empty exists on CompactString
     #[inline]
     pub fn len(&self) -> usize {
-        // the last byte stores our discriminant and stack length
-        let last_byte = self.last_byte();
+        // This ugly looking code results in two conditional moves and only one comparison, without
+        // branching. The outcome of a comparison is a tristate `{lt, eq, gt}`, but the compiler
+        // won't use this optimization if you match on `len_inline.cmp(&MAX_SIZE)`, so we have to
+        // do it manually.
 
-        // initially has the value of the stack length, conditionally becomes the heap length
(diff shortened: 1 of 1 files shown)


ParkMyCar:
This PR is great, thank you @Kijewski! Do you mind rebasing on `main`?

ParkMyCar:
Also this makes me wonder if storing the length as the last item of the heap variant would improve performance, i.e. `[ ptr | cap | len ]`. Then when reading the length the only load we'd need to do is the last word.

Separately I've been thinking about dropping support for 24 bytes and instead using just 23. Naively I think that would allow for a more performant implementation, at the cost of losing 1 byte, which might be worth it.

Kijewski:
> Also this makes me wonder if storing the length as the last item of the heap variant would improve performance, i.e. `[ ptr | cap | len ]`.

The problem is that the last byte cannot be used to store a value there, and we need to access the length more often than the capacity (I guess). If I'm not mistaken, the processor always reads two adjacent words at once since back in Pentium days, so I guess swapping the entries wouldn't be faster anyway. But I could easily be mistaken.

> Separately I've been thinking about dropping support for 24 bytes and instead using just 23.

I think 23 bytes is plenty for most applications. Or at least 24 is not much better than 23 bytes. I assume reading would same as fast, but writing to an inlined string could be quite a bit faster.

ParkMyCar:
Looks like we'll need to bump the MSRV to `1.59` support `std::arch::asm!`. Given that version is > 1 year old this seems okay to me

PR #284, opened 2023-08-10
feat: Impl `CompactString::from_str_to_{lower, upper}case`

NobodyXu:
Enable converting a `str` to lower/upper case and get a `CompactString` directly for inline storage.


NobodyXu:
MIPS CI failed due to pre-compiled libstd not available PowerPC Big Endian 32-bit failed due to `SIGILL: illegal instruction`, which I suppose is unrelated to this PR.

ParkMyCar:
Thank you!

PR #285, opened 2023-08-17
Fix and optimize `Capacity`

Kijewski:
In #276 I changed the value of the byte that tells if a string is stored on the heap, but I forgot to update the test for `Capacity`.

The compiler does not quite understand what we are doing in `Capacity::new()` and `Capacity::as_usize()`. Working on `usize` instead of `[u8; _]` allows the compiler to keep everything in registers instead of having to store the value on the stack to modify it.

Before:

```asm
Capacity::new:
	movabsq	$72057594037927935, %rcx
	andq	%rdi, %rcx
	movabsq	$-2882303761517117440, %rax
	orq	%rcx, %rax
	retq

Capacity::as_usize:
	movzwl	4(%rdi), %eax
	movzbl	6(%rdi), %ecx
	shll	$16, %ecx
	orl	%eax, %ecx
	shlq	$32, %rcx
	movl	(%rdi), %eax
	orq	%rcx, %rax
	retq
```

After:

```asm
Capacity::new:
	movabsq	$-2882303761517117440, %rax
	orq	%rdi, %rax
	retq

Capacity::as_usize:
	movabsq	$72057594037927935, %rax
	andq	%rdi, %rax
	retq
```

### DIFF
--- a/compact_str/src/repr/capacity.rs
+++ b/compact_str/src/repr/capacity.rs
@@ -3,20 +3,15 @@ use crate::repr::LastUtf8Char;
 // how many bytes a `usize` occupies
 const USIZE_SIZE: usize = core::mem::size_of::<usize>();
 
-/// Used to generate [`CAPACITY_IS_ON_THE_HEAP`]
-#[allow(non_snake_case)]
-const fn CAP_ON_HEAP_FLAG() -> [u8; USIZE_SIZE] {
-    // all bytes 255, with the last being HEAP_MASK
-    let mut flag = [255; USIZE_SIZE];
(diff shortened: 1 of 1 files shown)


Kijewski:
I wanted to add a CI test for i686, but locally I get always get ↓ right now even after deleting `target`:

> Inconsistency detected by ld.so: ./get-dynamic-info.h: 113: elf_get_dynamic_info: Assertion `info[DT_PLTREL]->d_un.d_val == DT_REL` failed!

ParkMyCar:
This is fantastic! 🔥🔥🔥🔥

PR #29, opened 2021-11-15
feature: Support no_std

ParkMyCar:

PR #287, opened 2023-08-26
support no_std

dragazo:
I have a compiler/interpreter project I'm developing for embedded systems, and this crate looks perfect for saving a ton of heap space and gaining locality for my many symbol tables. The only issue is that it needs to be `no_std` and #29 was closed a long time ago. So I've gone ahead and redone that work here.

It's pretty much all just a switch from `std` to `core` or `alloc` versions of things, plus a new (default) feature flag called `std` that enables a few small things like operations on `OsString` and the like.

The new version passes all test cases in standard mode, successfully builds in no-std mode, and miri has no issues with anything.

### DIFF
--- a/.github/workflows/nostd.yml
+++ b/.github/workflows/nostd.yml
@@ -1,0 +1,28 @@
+name: no-std
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+  workflow_dispatch:
+
+jobs:
+  no-std:
+    name: no-std build
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        opt: ["", "--release"]
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
+          override: true
+      - uses: actions-rs/cargo@v1
+        with:
+          use-cross: true
+          command: build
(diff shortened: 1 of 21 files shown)


dragazo:
Oh sure, I have a trick I use to test that in my no-std libs, which is to cross-compile targeting `thumb6m-none-eabi`, which does not (and likely will never) support the standard library. I just added a new github workflow for that.

The only issue it discovered was caused by a missing `LifetimeFree` impl for `String`, which is a no-std compatability issue in `castaway`. I've already fixed that in a fork, which is pending approval [here](https://github.com/sagebind/castaway/pull/13). Once that's merged and eventually published, we'll just need to switch out to the new `alloc` feature I added for that dep, and it should pass the CI I just added to this repo.

ParkMyCar:
> Oh sure, I have a trick I use to test that in my no-std libs, which is to cross-compile targeting `thumbv6m-none-eabi`, which does not (and likely will never) support the standard library. I just added a new github workflow for that.

Neat! That's a good trick to know

> The only issue it discovered was caused by a missing `LifetimeFree` impl for `String`, which is a no-std compatability issue in `castaway`. I've already fixed that in a fork, which is pending approval [here](https://github.com/sagebind/castaway/pull/13). Once that's merged and eventually published, we'll just need to switch out to the new `alloc` feature I added for that dep, and it should pass the CI I just added to this repo.

Sounds great, thanks for putting up a PR against `castaway`!

Issue #304, opened 2023-10-07
perf: Re-enable specialization for String when castaway is updated


ParkMyCar:
After merging #287 the specialization for `String` in the `ToString` trait was disabled, pending the release of this PR https://github.com/sagebind/castaway/pull/13.

PR #318, opened 2023-10-08
Implement diesel compatibility

Kijewski:
This PR allows using CompactString as representation for [diesel](https://diesel.rs/) text columns.


ParkMyCar:
Thank you for the PR! I'm going to release `v0.8.0-beta`, but we'll target to get trait impls for database crates in either a `v0.8.0-beta.1` or just `v0.8.0`, all dependent on https://github.com/ParkMyCar/compact_str/issues/304

tests/diesel/src/models.rs
```diff
@@ -0,0 +1,18 @@
...
+pub struct Todo {
+    pub id: i32,
+    pub title: CompactString,
+    pub done: bool,
+}
+
+#[derive(Insertable)]
+#[diesel(table_name = crate::schema::todos)]
+pub struct NewTodo<'a> {
+    pub title: &'a str,
```
NobodyXu, line 16: Why not make this

```suggestion
    pub title: CompactString,
```

to test serialization?

NobodyXu (changes requested):

tests/diesel/src/models.rs
Kijewski: Ah, yes, you're right!

tests/diesel/src/lib.rs
```diff
@@ -0,0 +1,81 @@
...
+use tempfile::tempdir;
+
+use crate::models::{
+    NewTodo,
+    Todo,
+};
+use crate::schema::todos;
+use crate::schema::todos::dsl::*;
+
+const TITLE: &str = "Say hello!";
```
NobodyXu, line 18: Could make it a `CompactString`:

```suggestion
const TITLE: CompactString = CompactString::from_static_str("Say hello!");
```

Kind of testing the static str impl and comparison with inline impl.

NobodyXu (changes requested):

tests/diesel/src/lib.rs
Kijewski: Good idea!

NobodyXu (approved):

ParkMyCar:
Overall this looks good to me, what do you think about moving `tests/diesel` to `examples/diesel` though? Reason being, that's where all of our current examples of how to use optional features live.

Also thanks to @NobodyXu for giving this a once over!

tests/diesel/src/lib.rs
```diff
@@ -0,0 +1,84 @@
...
+use crate::schema::todos::dsl::*;
+
+const TITLE: CompactString = CompactString::from_static_str("Say hello!");
+
+#[test]
+fn diesel_roundtrip() -> Result<(), Box<dyn std::error::Error>> {
+    // create and open a temporary database
+    let tempdir = tempdir()?;
+    let db_path = tempdir.path().join("compact_str.sqlite3");
+    let db_path = format_compact!("file://{}", db_path.to_str().unwrap());
```
ParkMyCar, line 28: Love the use of the macro!

Kijewski:
> what do you think about moving tests/diesel to examples/diesel though? Reason being, that's where all of our current examples of how to use optional features live.

I was not sure if how I am using diesel is best practice, and didn't want to call it an example because of that. But I guess it does not matter all that much. :) I move the code into `examples` and rebased the PR.

ParkMyCar (approved):

ParkMyCar:
Fantastic, thanks @Kijewski!

PR #321, opened 2023-10-09
Inline short static strings

Kijewski:
This way short static strings have better cache locality, too.


NobodyXu (approved):
This is awesome!

Inline <= 12/24B to avoid redirection.

NobodyXu:
But looks like the CI needs more work

ParkMyCar:
This looks good to me! Let's just get CI passing and then we can merge it

ParkMyCar:
Fantastic! Thanks @Kijewski!

PR #323, opened 2023-10-12
Make (re)allocations fallible

Kijewski:
The CI error in #321 is actually an implementation error. Once a CompactString gets so big that its capacity cannot be stored inline anymore, it will always be at least this big. This PR lets `shrink_to()` allocate new `Repr` with a capacity that can be inlined.

To facilitate this change I made allocations and re-allocations fallible. As an added bonus better panic messages will be generated, mentioning the calling locations of the failing (re)allocations. And the user gets two new methods, `try_with_capacity()` and `try_reserve`, to manually handle out-of-memory cases.

### DIFF
--- a/compact_str/src/lib.rs
+++ b/compact_str/src/lib.rs
@@ -206,8 +206,9 @@ impl CompactString {
     /// let boxed = CompactString::new(&b);
     /// ```
     #[inline]
+    #[track_caller]
     pub fn new<T: AsRef<str>>(text: T) -> Self {
-        CompactString(Repr::new(text.as_ref()))
+        CompactString(Repr::new(text.as_ref()).unwrap())
     }
 
     /// Creates a new inline [`CompactString`] at compile time.
@@ -310,8 +311,9 @@ impl CompactString {
     /// assert!(empty.is_heap_allocated());
     /// ```
     #[inline]
+    #[track_caller]
     pub fn with_capacity(capacity: usize) -> Self {
-        CompactString(Repr::with_capacity(capacity))
+        CompactString(Repr::with_capacity(capacity).unwrap())
     }
 
     /// Convert a slice of bytes into a [`CompactString`].
@@ -375,8 +377,9 @@ impl CompactString {
     /// ```
     #[inline]
     #[must_use]
+    #[track_caller]
     pub unsafe fn from_utf8_unchecked<B: AsRef<[u8]>>(buf: B) -> Self {
-        CompactString(Repr::from_utf8_unchecked(buf))
+        CompactString(Repr::from_utf8_unchecked(buf).unwrap())
     }
 
     /// Decode a [`UTF-16`](https://en.wikipedia.org/wiki/UTF-16) slice of bytes into a
@@ -536,8 +539,9 @@ impl CompactString {
     /// assert!(compact.capacity() >= 200);
     /// ```
     #[inline]
+    #[track_caller]
     pub fn reserve(&mut self, additional: usize) {
-        self.0.reserve(additional)
+        self.0.reserve(additional).unwrap()
     }
 
     /// Returns a string slice containing the entire [`CompactString`].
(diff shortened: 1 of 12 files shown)


compact_str/src/repr/iter.rs
```diff
@@ -19,7 +19,7 @@ impl FromIterator<char> for Repr {
         // If the size hint indicates we can't store this inline, then create a heap string
         let (size_hint, _) = iter.size_hint();
         if size_hint > MAX_SIZE {
-            return Repr::from_string(iter.collect(), true);
+            return Repr::from_string(iter.collect(), true).unwrap();
```
NobodyXu, line 22: Is there anyway to avoid that string creation?

compact_str/src/traits.rs
```diff
@@ -71,6 +71,7 @@ unsafe impl LifetimeFree for Repr {}
 ///     * For floats we use [`ryu`] crate which sometimes provides different formatting than [`std`]
 impl<T: fmt::Display> ToCompactString for T {
     #[inline]
+    #[track_caller]
     fn to_compact_string(&self) -> CompactString {
```
NobodyXu, line 75: Perhaps it's a good idea adding a try_ version of this fn?

NobodyXu (changes requested):

compact_str/src/repr/iter.rs
Kijewski: Yes, and the code is much easier on the eyes, IMO.

compact_str/src/repr/traits.rs
```diff
@@ -9,18 +9,22 @@ pub trait IntoRepr {
 }
 
 impl IntoRepr for f32 {
+    #[inline]
+    #[track_caller]
     fn into_repr(self) -> Repr {
         let mut buf = ryu::Buffer::new();
         let s = buf.format(self);
-        Repr::new(s)
+        Repr::new(s).unwrap()
```
NobodyXu, line 17: This should also be falliable since it's called in try_to_compact_string()

NobodyXu (changes requested):

compact_str/src/lib.rs
```diff
@@ -2392,4 +2427,88 @@ impl DoubleEndedIterator for Drain<'_> {
...
+    }
+}
+
+trait UnwrapWithMsg {
+    type T;
+
+    fn unwrap_with_msg(self) -> Self::T;
+}
+
+impl<T, E: fmt::Display> UnwrapWithMsg for Result<T, E> {
```
NobodyXu, line 2494: unwrap actually prints the error causing the panic https://doc.rust-lang.org/src/core/result.rs.html#1071-1073

NobodyXu (changes requested):

compact_str/src/lib.rs
Kijewski: `unwrap_failed()` uses `{error:?}`: <https://doc.rust-lang.org/src/core/result.rs.html#1651-1653>, which is not very descriptive in most cases.

compact_str/src/traits.rs
Kijewski: Yes, it is! Added.

compact_str/src/repr/traits.rs
Kijewski: You're right, fixed.

compact_str/src/lib.rs
Kijewski: With `.unwrap()` the message reads (in nightly):

> thread 'tests::test_alloc_excessively_long_string' panicked at src/tests.rs:1832:5:  
called `Result::unwrap()` on an `Err` value: ReserveError(())

With `.unwrap_with_msg()` the message reads:

> thread 'tests::test_alloc_excessively_long_string' panicked at src/tests.rs:1832:5:  
Cannot allocated memory to hold CompactString

Also the generated code is a tiny bit better, because the `error` is consumed by value, without a `&dyn` reference. This way no vtables or trampoline functions have to be generated.

NobodyXu: Thanks for explanation!

NobodyXu (approved):
LGTM, thanks for swift response to my review!

Kijewski:
> LGTM, thanks for swift response to my review!

And thank you for your reviews! :)

Issue #327, opened 2023-10-21
Consider implementing From<CompactString> for std::error::Error


tylerhawkes:
I was using this today and mostly able to just add `use compact_str::CompactStr as String;` in the imports of any file where I wanted to convert to CompactString. A couple places needed some `.into()` to convert strings like from `serde_json`, so not a big deal. However, in one place I was returning `Result<(), Box<dyn std::error::Error>>` which for String converts with a `.into()` call, but not for `CompactString`. 

The std library does it [here](https://doc.rust-lang.org/stable/std/string/struct.String.html#impl-From%3CString%3E-for-Box%3Cdyn+Error+%2B+Sync+%2B+Send,+Global%3E).

There are several other `From` impls that could also be done as well if they're missing.

Kijewski:
Sounds like a good idea. Maybe you could try to implement this conversion?

tylerhawkes:
Ok, I'll try to make a PR for this.

ParkMyCar:
Thanks for filing an issue @tylerhawkes! I love to see that we're getting closer to being a truly "drop-in" replacement for `String`. If you have the time to make a PR that would be great 😊

PR #328, opened 2023-10-23
Add missing From impls that String has

tylerhawkes:
Closes https://github.com/ParkMyCar/compact_str/issues/327

Each new impl has a corresponding test that requires that impl to be used for both long and short strings.

### DIFF
--- a/compact_str/src/lib.rs
+++ b/compact_str/src/lib.rs
@@ -2146,6 +2146,99 @@ impl<'a> From<&'a CompactString> for Cow<'a, str> {
     }
 }
 
+#[rustversion::since(1.60)]
+#[cfg(target_has_atomic = "ptr")]
+impl From<CompactString> for alloc::sync::Arc<str> {
+    fn from(value: CompactString) -> Self {
+        Self::from(value.as_str())
+    }
+}
+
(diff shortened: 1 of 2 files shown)


tylerhawkes:
It appears that `#[cfg(target_has_atomic = "ptr")]` was stabilized in 1.60. I tried adding it to fix the build failure for the `thumbv6m-none-eabi` target, but I can roll that back.

NobodyXu (approved):

compact_str/src/lib.rs
```diff
@@ -2130,6 +2130,98 @@ impl<'a> From<&'a CompactString> for Cow<'a, str> {
     }
 }
 
+#[cfg(target_has_atomic = "ptr")]
```
ParkMyCar, line 2133: Can you annotate this with `#[rustversion(since(1.60))]`? This way we can support more architectures and include this impl?

ParkMyCar (approved):
Thanks for adding these impls @tylerhawkes! Do you mind making the one `#[rustversion ... ]` change and re-basing on `main`? Once CI is green we'll merge!

ParkMyCar:
Sweeeeeeet, thanks again @tylerhawkes!

PR #329, opened 2023-10-23
Implement for sqlx

Kijewski:


Kijewski:
Mysql and Postgres are untested.

compact_str/src/features/sqlx/mysql.rs
```diff
@@ -0,0 +1,54 @@
...
+impl Type<MySql> for CompactString {
+    #[inline]
+    fn type_info() -> <MySql as Database>::TypeInfo {
+        <String as Type<MySql>>::type_info()
+    }
+}
+
+impl<'r> Decode<'r, MySql> for CompactString {
+    fn decode(value: MySqlValueRef<'r>) -> Result<Self, BoxDynError> {
+        let value = value.to_owned();
```
NobodyXu, line 33: Is it possible to avoid this `.to_owned()` call?

compact_str/src/features/sqlx/postgres.rs
```diff
@@ -0,0 +1,56 @@
+#![cfg_attr(docsrs, doc(cfg(feature = "sqlx-postgres")))]
```
NobodyXu, line 1: This file is mostly the same as mysql, maybe extract them as a macro?

compact_str/src/features/sqlx/sqlite.rs
```diff
@@ -0,0 +1,65 @@
...
+
+impl Type<Sqlite> for CompactString {
+    #[inline]
+    fn type_info() -> <Sqlite as Database>::TypeInfo {
+        <String as Type<Sqlite>>::type_info()
+    }
+}
+
+impl<'r> Decode<'r, Sqlite> for CompactString {
+    fn decode(value: SqliteValueRef<'r>) -> Result<Self, BoxDynError> {
```
NobodyXu, line 36: The decode impl is also similar to other two impls.

NobodyXu (changes requested):

compact_str/src/features/sqlx/mysql.rs
Kijewski: No needs, it's a reference counted type. Without .to_owned() you cannot do anything with it.

Kijewski:
I am using namespaced features in this PR now, which would mean that this MSRV is stuck to 1.60. Do you think that's okay?

compact_str/src/features/sqlx.rs
```diff
@@ -0,0 +1,97 @@
...
+                Encode::<'_, sqlx::$db>::encode_by_ref(&self.as_str(), buf)
+            }
+
+            impl_encode_common!($db);
+        }
+    };
+}
+
+impl_encode!("sqlx-mysql" => MySql);
+impl_encode!("sqlx-postgres" => Postgres);
```
ParkMyCar, line 80: Do you mind expanding these macros? Normally I would totally agree that using a macro here and adhering to DRY is the correct approach. But given we're not de-duplicating that much code, I don't believe the added complexity/overhead of using a macro with worth it

ParkMyCar:
> I am using namespaced features in this PR now, which would mean that this MSRV is stuck to 1.60. Do you think that's okay?

Yeah I think bumping the MSRV to 1.60 is a good idea! It will also help with https://github.com/ParkMyCar/compact_str/pull/328

ParkMyCar (commented):
Thanks for putting up this PR @Kijewski, I really appreciate it!

The implementation looks good, just one comment on the macro. Otherwise it looks like we'll need to update the [x-plat CI workflow](https://github.com/ParkMyCar/compact_str/blob/main/.github/workflows/cross_platform.yml#L82) and exclude at least the `sqlx/sqlite` feature since we're failing to build `sqlite` via `cross`.

Kijewski:
Yeah, I used a bit too many macros in here. :)

Could you have a look how to update the CI workflow? I don't know exactly how to exclude the test when cross compiling.

compact_str/src/features/sqlx.rs
```diff
@@ -0,0 +1,113 @@
...
+where
+    DB: Database,
+    for<'x> &'x str: Decode<'x, DB> + Type<DB>,
+{
+    fn decode(value: <DB as HasValueRef<'r>>::ValueRef) -> Result<Self, BoxDynError> {
+        let value = value.to_owned();
+        let value: &str = value.try_decode()?;
+        Ok(value.try_to_compact_string()?)
+    }
+}
```
tylerhawkes, line 49: Is there a reason to not specifically implement Decode for each of the Database variants? They each look like they have faster code paths for decoding `&str` that can be used to create a CompactStr.

ParkMyCar:
Thanks for the PR @Kijewski!

The CI issue looks to be `sqlx-sqlite` trying to build `libsqlite3-sys`, which not surprisingly fails in `cross`, I ignored the `sqlx-*` features for x-plat builds with cross. The MSRV check fails because it looks like proptest updates their MSRV again, I'll fix this in a separate PR though!

PR #330, opened 2023-10-24
Simplify `is_empty()`

Kijewski:
Before:

```asm
0000000000000000 <<compact_str::CompactString>::is_empty>:
   0:   48 8b 77 08             mov    0x8(%rdi),%rsi
   4:   0f b6 57 17             movzbl 0x17(%rdi),%edx
   8:   48 8d 8a 40 ff ff ff    lea    -0xc0(%rdx),%rcx
   f:   48 83 f9 18             cmp    $0x18,%rcx
  13:   b8 18 00 00 00          mov    $0x18,%eax
  18:   48 0f 42 c1             cmovb  %rcx,%rax
  1c:   48 81 fa d8 00 00 00    cmp    $0xd8,%rdx
  23:   48 0f 43 c6             cmovae %rsi,%rax
  27:   48 85 c0                test   %rax,%rax
  2a:   0f 94 c0                sete   %al
  2d:   c3                      ret
```

After

```asm
0000000000000000 <<compact_str::CompactString>::is_empty>:
   0:   48 8b 57 08             mov    0x8(%rdi),%rdx
   4:   0f b6 4f 17             movzbl 0x17(%rdi),%ecx
   8:   48 8d 81 40 ff ff ff    lea    -0xc0(%rcx),%rax
   f:   48 81 f9 d8 00 00 00    cmp    $0xd8,%rcx
  16:   48 0f 43 c2             cmovae %rdx,%rax
  1a:   48 85 c0                test   %rax,%rax
  1d:   0f 94 c0                sete   %al
  20:   c3                      ret
```

### DIFF
--- a/compact_str/src/tests.rs
+++ b/compact_str/src/tests.rs
@@ -1853,95 +1853,25 @@ fn multiple_niches_test() {
 
 #[test]
 fn test_is_empty() {
-    // 0
+    const ZEROS: &[&str] = &[
+        "\0",                                                 // 1
+        "\0\0\0\0\0\0\0\0\0\0\0\0",                           // 12
+        "\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0",   // 24
+        "\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0", // 25
+    ];
+
     assert!(CompactString::new("").is_empty());
     assert!(CompactString::from_static_str("").is_empty());
-    // 1
-    assert!(!CompactString::new("\0").is_empty());
-    assert!(!CompactString::from_static_str("\0").is_empty());
-    // 2
-    assert!(!CompactString::new("\0\0").is_empty());
-    assert!(!CompactString::from_static_str("\0\0").is_empty());
-    // 3
-    assert!(!CompactString::new("\0\0\0").is_empty());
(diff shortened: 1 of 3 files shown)


Kijewski:
3 instructions less, and 1 less register used. Not as much of an optimization as I had hoped for.

NobodyXu (approved):

compact_str/src/tests.rs
```diff
@@ -1850,3 +1850,98 @@ fn multiple_nieches_test() {
         core::mem::size_of::<String>()
     );
 }
+
+#[test]
+fn test_is_empty() {
```
mcronce, line 1855: It might be worth simplifying this test with `<&str>::repeat()`, something like

```rust
assert!(CompactString::new("").is_empty());
assert!(CompactString::from_static_str("").is_empty());

for i in 1..=25 {
  let s = Box::leak("\0".repeat(i).into_boxed_str());
  assert!(!CompactString::new(s).is_empty());
  assert!(!CompactString::from_static_str(s).is_empty());
}
```

mcronce: Never mind - just realized `CompactString::from_static_str()` is a `const fn` - that won't work with `&'static str`s you get by leaking at runtime

Kijewski: I simplified the test a bit.

ParkMyCar:
Interesting! Thanks for the PR, and for abstracting out `ensure_read(...)`!

PR #282, opened 2023-07-22
Improve CI: Rm use of deprecated actions

NobodyXu:
Fixed #280 and fixed #279

also:
 - Bump `actions/checkout` to v4
 - use `taiki-e/install-action` to install any binary crates
 - use `Swatinem/rust-cache@v2` for caching
 - use `strategy` to avoid similar/duplicate jobs
 - pass `--no-deps` to `cargo-clippy`
 - Use `cargo-nextest` for `randomize-layout`
 - Run doc test in job `test`
 - Use `cargo-nextest` to speedup `miri`
 - Run `cargo-check` instead of `cross-build` in `nostd.yml` since
   there's no need for `cross`

Signed-off-by: Jiahao XU <Jiahao_XU@outlook.com>

PR #331, opened 2023-10-25
fix: Merge skew and MSRV workflow

ParkMyCar:
With some recent PRs we hit some merge skews that prevented a test from compiling, this PR fixes those issues. It also bumps the MSRV CI workflow to 1.60, suddenly when running in 1.59 we get the following error:

```
$ cargo +1.59 check --no-default-features                                                                                                                         [±fix/merge-skew ●]
    Updating crates.io index
error: failed to select a version for `rkyv`.
    ... required by package `compact_str v0.8.0-beta (/Users/parker/Development/compact_str/compact_str)`
versions that meet the requirements `^0.7` are: 0.7.41, 0.7.40, 0.7.39, 0.7.38, 0.7.37, 0.7.36, 0.7.35, 0.7.34, 0.7.33, 0.7.32, 0.7.31, 0.7.30, 0.7.29, 0.7.28, 0.7.27, 0.7.26, 0.7.25, 0.7.24, 0.7.23, 0.7.22, 0.7.21, 0.7.20, 0.7.19, 0.7.18, 0.7.17, 0.7.16, 0.7.15, 0.7.14, 0.7.13, 0.7.12, 0.7.11, 0.7.10, 0.7.9, 0.7.8, 0.7.7, 0.7.6, 0.7.5, 0.7.4, 0.7.3, 0.7.2, 0.7.1, 0.7.0

all possible versions conflict with previously selected packages.

  previously selected package `rkyv v0.7.31`
    ... which satisfies dependency `rkyv = "^0.7"` of package `compact_str v0.8.0-beta (/Users/parker/Development/compact_str/compact_str)`

failed to select a version for `rkyv` which could resolve this conflict
```
... which is confusing, since the version of `rkyv` selected satisfies the dependency, and there haven't been any new versions of `rkyv` released in 5 months? It seems related to this issue, https://github.com/rust-lang/cargo/issues/10189, which indicates something further down in the dependency chain is to blame. Haven't had a chance to dig further, so I bumped the MSRV workflow to 1.60

### DIFF
--- a/.github/workflows/msrv.yml
+++ b/.github/workflows/msrv.yml
@@ -42,7 +42,7 @@ jobs:
         # Note2: Even though our MSRV is 1.59, we only test from 1.60 because we hit the issue
         # described in <https://github.com/rust-lang/cargo/issues/10189> when using 1.59.
         run: |
-          cargo hack check --features bytes,markup,quickcheck,rkyv,serde,smallvec --version-range 1.60..
+          cargo hack check --features bytes,markup,quickcheck,rkyv,serde,smallvec --manifest-path=compact_str/Cargo.toml --version-range 1.60..
           cargo hack check --features bytes,markup,quickcheck,rkyv,serde,smallvec,proptest,arbitrary --manifest-path=compact_str/Cargo.toml --version-range 1.64..
 
   feature_powerset:
--- a/.github/workflows/msrv.yml
+++ b/.github/workflows/msrv.yml
@@ -30,16 +30,19 @@ jobs:
       - uses: Swatinem/rust-cache@v2
 
       - name: cargo test msrv..
-        # note: we exclude both `arbitrary` and `proptest` from here because their MSRVs were both
+        # Note: we exclude both `arbitrary` and `proptest` from here because their MSRVs were both
         # bumped on a minor version release:
         #
         # - abitrary >= 1.1.14 has an MSRV >= 1.63
         # - proptest >= 1.1.0 has an MSRV >= 1.60
         #
         # Instead of pinning to a specific version of `arbitrary` or `proptest`, we'll let user's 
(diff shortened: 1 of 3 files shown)


NobodyXu:
@ParkMyCar would be great if you can merge my PR #282

.github/workflows/msrv.yml
```diff
@@ -30,16 +30,19 @@ jobs:
       - uses: Swatinem/rust-cache@v2
```
NobodyXu, line 31: There's one way to fix this without bumping MSRV:

Before the line rust-cache, do a

```
cargo +nightly update -Zminimal
```

That will guarantee that the version specified in our `Cargo.toml` is used.

It needs to be done before rust-cache, to make sure it picks up the Cargo.lock generated and use it when computing cache key.

NobodyXu (changes requested):

.github/workflows/msrv.yml
ParkMyCar: Thanks for the pointer! Just added this, hopefully it works 🤞

Kijewski:
You actually don't need to install a nightly toolchain, just set `RUSTC_BOOTSTRAP=1`. Compare: https://github.com/strawlab/iana-time-zone/blob/main/.github/workflows/rust.yml#L111-L138

Kijewski:
1.60 added [namespaced features and weak dependencies](https://doc.rust-lang.org/cargo/reference/unstable.html#weak-dependency-features), which might come in handy if we want to add more integrations without a semver breaking change.

And since 1.60 was released on [2022-04-07](https://caniuse.rs/versions/1.60), I guess we should not invest any more work into making 1.59 work.

Kijewski (approved):

PR #336, opened 2023-10-29
Add `const_new()`; remove `new_inline()` and `from_static_str()`

### LINKED ISSUE #322: Rethink `new_inline()` and `from_static_str()`
In my opinion, `from_static_str()` should be the default option for a user to create a const or static CompactString. This way they don't have to think about how CompactStrings actually work. Things just work.

1. I would rename `from_static_str()` into `const_new()` (borrowing the name e.g. from tokio), to make it obvious that this is the function one should use.
2. I would rename `new_inline()` into `try_new_inline()` and make this fallible.

What do you think? I know that this is a big change, so I would keep `new_inline()` with a deprecation notice around for a while.

Kijewski:
Closes #322.

Please see #322 for an explanation.

### DIFF
--- a/compact_str/src/repr/mod.rs
+++ b/compact_str/src/repr/mod.rs
@@ -93,7 +93,7 @@ impl Repr {
     }
 
     #[inline]
-    pub const fn from_static_str(text: &'static str) -> Self {
+    pub const fn const_new(text: &'static str) -> Self {
         if text.len() <= MAX_SIZE {
             let inline = InlineBuffer::new_const(text);
             Self::from_inline(inline)
@@ -839,7 +839,7 @@ mod tests {
         assert_eq!(repr.len(), s.len());
 
         // test StaticStr variant
-        let repr = Repr::from_static_str(s);
+        let repr = Repr::const_new(s);
         assert_eq!(repr.as_str(), s);
         assert_eq!(repr.len(), s.len());
     }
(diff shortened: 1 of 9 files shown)


compact_str/src/repr/mod.rs
```diff
@@ -452,6 +439,36 @@ impl Repr {
...
+        if self.last_byte() >= HEAP_MASK {
+            pointer = heap_pointer;
+        }
+        pointer
+    }
+
+    /// Return a mutable reference to the underlying buffer
+    ///
+    /// # Safety
+    /// Must only be called on a dynamically allocated [`Repr`].
```
NobodyXu, line 457: This should be "not static repr", since it only asserts on non-static repr

NobodyXu (changes requested):

compact_str/src/repr/mod.rs
Kijewski: My reasoning was flawed anyway. I changed `as_mut_ptr()`'s implementation, so it will automatically allocated writable data if it previously contained immutable data.

NobodyXu (approved):

ParkMyCar:
Hey @Kijewski! Thanks again for a great PR!

There seem to be some changes include here that are not related to adding `const_new(...)` and deprecating `new_inline(...)` and `from_static_str(...)`, e.g. the ptr APIs and some trait impls. Those look like great changes, but do you mind splitting them into a separate PR?

Kijewski:
Yeah, that were minor points that I found while walking through the code to do the renaming. I moved it into #350.

ParkMyCar:
Thanks for separating out the changes!

PR #342, opened 2023-10-31
 On 64bit platforms, the capacity is always stored inline 

Kijewski:
On 64 bit platforms we do not have to check if the capacity is stored on the heap, because it will never be stored on the heap.

### DIFF
--- a/compact_str/src/repr/capacity.rs
+++ b/compact_str/src/repr/capacity.rs
@@ -23,6 +23,7 @@ const HEAP_MARKER: usize = {
 ///
 /// All bytes `255`, with the last being [`LastUtf8Char::Heap`], using the same amount of bytes
 /// as `usize`. Example (64-bit): `[255, 255, 255, 255, 255, 255, 255, 216]`
+#[cfg(not(target_pointer_width = "64"))]
 const CAPACITY_IS_ON_THE_HEAP: Capacity = Capacity(VALID_MASK | HEAP_MARKER);
 
 /// The maximum value we're able to store, e.g. on 64-bit arch this is 2^56 - 2.
@@ -100,6 +101,7 @@ impl Capacity {
     /// Returns whether or not this [`Capacity`] has a value that indicates the capacity is being
     /// stored on the heap
     #[inline(always)]
+    #[cfg(not(target_pointer_width = "64"))]
     pub fn is_heap(self) -> bool {
         self == CAPACITY_IS_ON_THE_HEAP
     }
--- a/compact_str/src/repr/heap.rs
+++ b/compact_str/src/repr/heap.rs
@@ -2,6 +2,7 @@ use core::alloc::Layout;
 use core::{
     cmp,
     mem,
+    ops,
     ptr,
 };
 
@@ -89,6 +90,7 @@ impl HeapBuffer {
     /// Return the capacity of the [`HeapBuffer`]
     #[inline]
     pub fn capacity(&self) -> usize {
+        #[cfg(not(target_pointer_width = "64"))]
         #[cold]
         fn read_capacity_from_heap(this: &HeapBuffer) -> usize {
             // re-adjust the pointer to include the capacity that's on the heap
@@ -101,12 +103,13 @@ impl HeapBuffer {
             usize::from_ne_bytes(buf)
         }
 
+        #[cfg(not(target_pointer_width = "64"))]
         if self.cap.is_heap() {
-            read_capacity_from_heap(self)
-        } else {
-            // SAFETY: Checked above that the capacity is on the stack
-            unsafe { self.cap.as_usize() }
+            return read_capacity_from_heap(self);
         }
+
+        // SAFETY: Checked above that the capacity is on the stack
+        unsafe { self.cap.as_usize() }
     }
 
     /// Try to grow the [`HeapBuffer`] by reallocating, returning an error if we fail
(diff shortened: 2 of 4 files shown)


Kijewski:
Interesting. This does not panic in release mode:

```rust
fn test_alloc_excessively_long_string() {
    CompactString::with_capacity((1 << 56) - 2);
}
```

But this does:

```rust
fn test_alloc_excessively_long_string() {
    let x = CompactString::with_capacity((1 << 56) - 2);
    dbg!(x.as_ptr());
}
```

Can it be that the compiler understood that we are doing nothing in the function body, so it is culled entirely?

ParkMyCar:
Thanks for this PR @Kijewski! We actually used to do this, had a bunch of conditional compilation specifically for 64-bit architectures. It was quite complicated which is why I got rid of it.

On another note though, what I would like to do is get rid of this capacity inline vs capacity on the heap distinction all together! I think what we can do is:

* `0XXX_XXXX` - Inline packed
* `10XX_XXXX` - Inline packed
* `11XX_XXXX` - Inline with length
* `111X_XXXX` - Heap

There is some trickiness with this approach in supporting the Option niche optimization, and static strings, but I believe supporting both of these is still possible. Now on 32-bit architectures this means the maximum heap based string we can support is ~512MiB, but IMO this trade-off is worth it for the (what I believe will be) cleaner codebase and improved performance. What are your thoughts?

Kijewski:
Actually, I don't think that the code is much more difficult to read with a few `cfg(target_pointer_width = "64")` or `cfg(not(target_pointer_width = "64"))` here and there. A few could guards could be removed:

```rust
#[cfg(not(target_pointer_width = "64"))]
#[cold]
fn handle_unlikely_32bit_only_case(this: &HeapBuffer) -> usize {
    …
}

#[cfg(not(target_pointer_width = "64"))]
if self.is_unlikely_32bit_only_case() {
    return handle_unlikely_32bit_only_case(self);
}
```

→


```rust
#[cfg(not(target_pointer_width = "64"))]
if self.is_unlikely_32bit_only_case() {
    #[cold]
    fn handle_unlikely_32bit_only_case(this: &HeapBuffer) -> usize {
        …
    }
    return handle_unlikely_32bit_only_case(self);
}
```

I only didn't do that because it would make the diff even bigger.

----

But I think containing some bits of the capacity in last byte would work, and there would still be 8 niches left (0b110x_xxxx = 5 bits = 32, -1 for static strings, -23 for the lengths).

I am unsure what the performance implications would be. Spontaneously I'd guess that it would benefit 32 bit systems, but it could hurt 64 bit systems for the likely read-mostly use case.

Kijewski:
Does not seem to work. If you have an `#[repr(u8)] enum { A = 1, D = 4 }`, then the values 2 and 3 are not considered as niches. Adding `B(Infallible) = 1, B(Infallible) = 2` does not work either.

I think this PR is useful anyway, (at least until storing the capacity is modified). IMO it does not make understanding the current implementation more difficult, and it looks like the compiler can do a few more optimizations with the change.

overlookmotel:
I believe if you switch around the bit patterns for inline and heap, you can move the unused values to the end of the range, where they should work as niches. 248-255 = 8 niches.

* 0XXX_XXXX - Inline packed
* 10XX_XXXX - Inline packed
* 110X_XXXX - Heap
* 111X_XXXX - Inline with length

As you say, that leaves 29 bits for capacity on 32-bit systems = 512 MiB.

overlookmotel:
Actually, maybe this is quite good:

* `0XXX_XXXX` - Inline packed
* `10XX_XXXX` - Inline packed
* `110X_XXXX` - Heap
* `1101_1111` - Static string
* `111X_XXXX` - Inline with length

Heap capacity on 32-bit is capped to 0x1EFFFFFF (496 MiB minus 1).

That leaves 0x1F in last byte free to be the flag for static string.

i.e. Possible values for last byte:

* 0 - 191: Inline packed
* 192 - 222: Heap
* 223: Static string
* 224 - 247: Inline with length
* 247 - 255: Unused (9 x niches)

Can reuse the intermediate value used in calculating inline length, so *hopefully* will produce no extra instructions:

Current:

```rs
const LENGTH_MASK: u8 = 192;
const HEAP_MASK: u8 = 224;
const MAX_SIZE: usize = 12; // 32-bit

pub fn len(&self) -> usize {
  let len_heap = ensure_read(self.1);
  let last_byte = self.last_byte();

  let mut len = (last_byte as usize)
    .wrapping_sub(LENGTH_MASK as usize)
    .min(MAX_SIZE);

  if last_byte >= HEAP_MASK {
    len = len_heap;
  }
  len
}
```

After change:

```rs
const LENGTH_MASK: u8 = 224;
const HEAP_MASK: u8 = 192;
const HEAP_MASK_AFTER_SUB: usize = (HEAP_MASK as usize)
  .wrapping_sub(LENGTH_MASK as usize);
const MAX_SIZE: usize = 12; // 32-bit
  
pub fn len(&self) -> usize {
  let len_heap = ensure_read(self.1);
  let last_byte = self.last_byte();

  let mut len = (last_byte as usize)
    .wrapping_sub(LENGTH_MASK as usize);
  let is_heap = len >= HEAP_MASK_AFTER_SUB;
  len = len.min(MAX_SIZE);

  if is_heap {
    len = len_heap;
  }
  len
}
```

PR #358, opened 2024-01-07
Align InlineBuffer same as Repr

### LINKED ISSUE #353: Should `InlineBuffer` be aligned same as `CompactStr`?
`InlineBuffer` is created for inline strings, and then transmuted into a `CompactStr`.

```rs
#[repr(transparent)]
pub struct InlineBuffer(pub [u8; MAX_SIZE]);
```

`InlineBuffer` has alignment of 1, whereas `CompactStr` has alignment of 8 (on 64-bit).

[mem::transmute](https://doc.rust-lang.org/std/mem/fn.transmute.html) does not have requirement that source and destination have same alignment, but nonetheless I'd have expected it could produce more optimized ASM (aligned reads) if they do.

However, in this test I can't see any difference in ASM between aligned and unaligned versions: https://godbolt.org/z/j5vGTxWxa

So probably I'm wrong about this. But if I am, would be interested to know why!

Incidentally, there is 1 unaligned move `movups` instruction in the ASM. I'd have expected all instructions could be aligned (i.e. `movaps`).

overlookmotel:
Fixes #353.

I also added assertions that `HeapBuffer` and `StaticStr` have same alignment as `Repr`.

### DIFF
--- a/compact_str/src/repr/heap.rs
+++ b/compact_str/src/repr/heap.rs
@@ -39,6 +39,7 @@ pub struct HeapBuffer {
     pub cap: Capacity,
 }
 static_assertions::assert_eq_size!(HeapBuffer, Repr);
+static_assertions::assert_eq_align!(HeapBuffer, Repr);
 
 impl HeapBuffer {
     /// Create a [`HeapBuffer`] with the provided text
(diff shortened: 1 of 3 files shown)


overlookmotel:
Any idea why CI is failing on MacOS?

```
could not compile `diesel_derives` (lib)
```

I am *guessing* that's unrelated to content of this PR?

NobodyXu (approved):

NobodyXu:
Yeah I think the CI failure is due to the dep.

PR #370, opened 2024-03-10
Some slightly performance & style updates

yhx-12243:
* In the function `fn as_bytes` in `impl CompactString`, there is an redundant slicing (with a `self.len()` call) which is unnecessary after the update c95edd5.
* This crate has documented that for 64-bit architectures the capacity will be always inlined. So the logic of `is_heap` can be replaced with directly returning `false` to reduce some code generation.

### DIFF
--- a/compact_str/src/lib.rs
+++ b/compact_str/src/lib.rs
@@ -614,7 +614,7 @@ impl CompactString {
     /// ```
     #[inline]
     pub fn as_bytes(&self) -> &[u8] {
-        &self.0.as_slice()[..self.len()]
+        self.0.as_slice()
     }
 
     // TODO: Implement a `try_as_mut_slice(...)` that will fail if it results in cloning?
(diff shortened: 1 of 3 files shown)


yhx-12243:
The test seems to be changed, it's likely that the compiler optimized out the whole test and did not allocate memory at all.
https://rust.godbolt.org/z/qc349PrGv

yhx-12243:
May be we should wrap this test in a something like `core::hint::black_box`, or maybe do other things to make sure the compiler will not optimize out the "unused" allocation of long string.

yhx-12243:
Hello @ParkMyCar , do you have time to review this PR if you're spared?

ParkMyCar:
Geez @yhx-12243 this sat for so long, I'm sorry about that! Thanks for the PR and the performance improvements, they look good to me!

PR #371, opened 2024-03-31
Add: CompactCowStr

aobatact:
This pull request aims to add a new struct called `CompactCowStr` to incorporate Cow for `CompactString`. 
This proposal has been previously discussed in issue #223.

## Question

- Reuse of `Repr`:
    `CompactCowStr` reuses `Repr`, which may potentially impact `CompactString`.
    If `Cow` flag of `LastUtf8Char` leaked to `CompactString`, this might cause UB.

- Methods:
    Which methods equivalent to those in `CompactString` are necessary for `CompactCowStr`?

- Inlining:
    For the new method, should we inline it?

-  Name Discussion:
    Previous pull request suggested the name `CompactCow`, but considering its relationship to String, I have changed it to `CompactCowStr`.

### DIFF
--- a/compact_str/src/cow.rs
+++ b/compact_str/src/cow.rs
@@ -1,0 +1,1532 @@
+use core::hash::{Hash, Hasher};
+use core::ops::{Add, AddAssign, RangeBounds};
+use core::str::FromStr;
+use core::{
+    borrow::{Borrow, BorrowMut},
+    cmp::Ordering,
+    marker::PhantomData,
+    ops::{Deref, DerefMut},
+    str::Utf8Error,
+};
+
+use alloc::boxed::Box;
+use alloc::fmt;
+use alloc::{borrow::Cow, string::String};
+
+use crate::Drain;
+use crate::{repr::Repr, CompactString, ReserveError, UnwrapWithMsg, Utf16Error};
+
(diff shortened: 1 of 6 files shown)


aobatact:
miri is failing with "unknown unstable option: `miri-check-number-validity`".
It seems the CI needs to be fixed?

NobodyXu:
I think that miri cmdline option might be removed in the nightly, so it probably needs to also be removed from CI?

aobatact:
Fixed CI in #374

ParkMyCar:
Hey @aobatact, sorry for the delay and thank you for the PR! I took a quick look and I think we're on the right path, but I've paged out a decent amount of the context here. I have a long train ride this weekend which I'll use to dive in deep here.

jf2048:
Is there some reason this needs to reuse `Repr` with its own niche? A simple enum implementation seems to have the same layout.

```rust
pub enum CompactCowStr<'a> {
    Borrowed(&'a str),
    Owned(CompactString),
}
assert_eq!(size_of::<CompactCowStr>(), size_of::<CompactString>());
assert_eq!(size_of::<Option<CompactCowStr>>(), size_of::<CompactString>());
```

PR #373, opened 2024-04-29
fix: remove -Zmiri-check-number-validity from ci

aobatact:
PR #371 is failing in the CI due to the utilization of the removed MIRI flag `-Zmiri-check-number-validity`. This PR remove the flag from CI.

See for https://github.com/rust-lang/miri/pull/2151 the removal.


ParkMyCar:
Thanks @aobatact!

PR #379, opened 2024-05-07
Remove invalid UTF8 check from `Repr::from_utf8_unchecked`

overlookmotel:
This PR removes the following code from `Repr::from_utf8_unchecked`:

https://github.com/ParkMyCar/compact_str/blob/d4798639c24cacf528770d83e8842336b79fbc39/compact_str/src/repr/mod.rs#L126-L137

In my opinion, this code should be removed for 2 reasons:

1. It is part of the safety contract for `from_utf8_unchecked` that `buf` is the bytes of a valid UTF8 string. This check penalizes users who follow the safety contract (which everyone should), in order to support users who don't (and therefore have opted in to UB).
2. This check alone is not enough to avoid UB anyway. If `buf` ends with the first byte of a multi-byte Unicode character, for example calling `.as_str().chars().collect::<Vec<char>>()` on the resulting `CompactString` will cause an out of bounds read.

### DIFF
--- a/compact_str/src/repr/mod.rs
+++ b/compact_str/src/repr/mod.rs
@@ -123,19 +123,6 @@ impl Repr {
         // Create a Repr with enough capacity for the entire buffer
         let mut repr = Repr::with_capacity(bytes_len)?;
 
-        // There's an edge case where the final byte of this buffer == `HEAP_MASK`, which is
-        // invalid UTF-8, but would result in us creating an inline variant, that identifies as
-        // a heap variant. If a user ever tried to reference the data at all, we'd incorrectly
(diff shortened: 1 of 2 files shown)


overlookmotel:
CI fails are I believe caused by tests which are breaking the safety constraint of `utf8_unchecked`. If intention is to merge this, those tests should be removed/adapted.

ParkMyCar:
Sorry for the delay here @overlookmotel, and thank you for the PR!

This is a very good point, I would like to add more documentation explaining what might occur if the user does not provide valid UTF-8. This PR has sat open for long enough though, I'll merge as-is and add the docs myself :)

overlookmotel:
No worries at all. Open source! Everyone is busy...

PR #393, opened 2024-06-24
Add `borsh` support

SmnTin:
This PR adds support for the `borsh` serialization library, similar to that of `serde` and `rkyv`.

### DIFF
--- a/compact_str/src/features/borsh.rs
+++ b/compact_str/src/features/borsh.rs
@@ -1,0 +1,78 @@
+#![cfg_attr(docsrs, doc(cfg(feature = "borsh")))]
+
+use borsh::io::{
+    Error,
+    ErrorKind,
+    Read,
+    Result,
+    Write,
+};
+use borsh::{
+    BorshDeserialize,
+    BorshSerialize,
+};
+
+use crate::CompactString;
+
+impl BorshSerialize for CompactString {
+    fn serialize<W: Write>(&self, writer: &mut W) -> Result<()> {
+        self.as_str().serialize(writer)
+    }
+}
+
+impl BorshDeserialize for CompactString {
+    fn deserialize_reader<R: Read>(reader: &mut R) -> Result<Self> {
+        let len = u32::deserialize_reader(&mut *reader)? as usize;
+
+        // Do not call `Error::new` on OOM as it allocates.
+        let mut s = CompactString::try_with_capacity(len).map_err(|_| ErrorKind::OutOfMemory)?;
+
+        // SAFETY: The current length is zero so the bytes are uninterpreted
+        // and don't have to form valid UTF-8
+        let buf = unsafe { s.as_mut_bytes() };
+
+        reader.read_exact(&mut buf[..len])?;
+        std::str::from_utf8(&buf[..len]).map_err(|err| Error::new(ErrorKind::InvalidData, err))?;
+
+        // SAFETY: The first `len` bytes are initialized and are valid UTF-8
+        unsafe { s.set_len(len) };
+
+        Ok(s)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use alloc::string::String;
+
+    use test_strategy::proptest;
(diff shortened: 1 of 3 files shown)


compact_str/src/features/borsh.rs
```diff
@@ -0,0 +1,94 @@
...
+            let mut buf = [0u8; MAX_SIZE];
+            reader.read_exact(&mut buf[..len])?;
+            let s = str::from_utf8(&buf[..len])
+                .map_err(|err| Error::new(ErrorKind::InvalidData, err))?;
+            Ok(CompactString::from(s))
+        } else {
+            let mut buf = Vec::new();
+            buf.try_reserve_exact(len)?;
+            buf.resize(len, 0u8);
+            reader.read_exact(&mut buf[..])?;
```
NobodyXu, line 42: With the following implementation, we can take advantages of `read_buf` on stable rust, so that it doesn't have to `memset` everything to zero before reading:

```suggestion
            reader.take(len as u64).read_to_end(&mut buf)?;
```

compact_str/src/features/borsh.rs
```diff
@@ -0,0 +1,94 @@
...
+        const VALUE: &str = "Hello, 🌍!";
+
+        let bytes_compact = borsh::to_vec(&CompactString::from(VALUE)).unwrap();
+        let bytes_control = borsh::to_vec(&String::from(VALUE)).unwrap();
+        assert_eq!(&*bytes_compact, &*bytes_control);
+
+        let compact: CompactString = borsh::from_slice(&bytes_compact).unwrap();
+        let control: String = borsh::from_slice(&bytes_control).unwrap();
+        assert_eq!(compact, VALUE);
+        assert_eq!(control, VALUE);
```
NobodyXu, line 73: This looks mostly identical to `proptest_roundtrip`, I think you could extract that to a new function?

compact_str/src/features/borsh.rs
SmnTin: Neat!

SmnTin: I see two issues with this, however:
- `read_to_end` might read less than `len` if there is less. I added a failing test and a fix, and it looks a bit ugly now
- I am not sure about the performance of `read_to_end`. It is optimized not to over-allocate when we don't know the exact size, but we know the exact size. Yes, it has special implementations for many standard implementors, but the performance gains might be seen only when reading huge strings.

NobodyXu: That's indeed an issue...

So it's unfortunate read_buf hasn't stablised yet, the only way to avoid the resize without `.take(n).read_to_end(&mut vec)` is to use [`Vec::spare_capacity_mut`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.spare_capacity_mut), but it contains unsafe and will be UB, if the reader actually reads from the uninitialised buffer, instead of just writing to it.

SmnTin: Well, I thought a bit. `memset`-ing should be pretty fast on modern machines. We should be more concerned about DoS attacks if the requested length is too big. E.g. what Borsh's `u8::vec_from_reader` [does](https://docs.rs/borsh/1.5.1/src/borsh/de/mod.rs.html#156-184), which is [used](https://docs.rs/borsh/1.5.1/src/borsh/de/mod.rs.html#400-422) in the implementation for `Vec<T>`. Unfortunately, even though this method is public, it is hidden, so the users aren't supposed to use it. We can just copy the code though :)

NobodyXu: Thanks, it's the best we can do for now

SmnTin: There is nothing else to do then

NobodyXu (approved):
Thanks, LGTM!

ParkMyCar:
Hey @SmnTin thanks for the PR! And thanks @NobodyXu for the review!

PR #399, opened 2024-07-13
impl sqlx::postgres::PgHasArrayType

imbolc:
This impl is necessary to use `CompactString` in newtypes


imbolc:
The failing check doesn't seem to relate to the changes

ParkMyCar:
Ooof sorry this was missing, thank you for the PR!

PR #401, opened 2024-08-18
Document CompactString has niches

stepancheg:
Add `assert_eq_size!` assertion.


ParkMyCar:
Good call, thanks for the PR!

Issue #407, opened 2024-09-13
[RUSTSEC-2024-0363] Support sqlx 0.8


rshigapov-bhft:
sqlx 0.8 has fixed [RUSTSEC-2024-0363.](https://rustsec.org/advisories/RUSTSEC-2024-0363.html)

I see two ways to support it:

* drop sqlx 0.7 support and implement support for sqlx 0.8
* add an extra feature like "sqlx-07" that allows users to keep working with sqlx 0.7 until they migrate to sqlx 0.8

ParkMyCar:
Thanks for the heads up! I'll publish a `v0.8.1` and `v0.7.2` that has a `sqlx_08` feature, which should be semver compatible and allow folks to opt into sqlx 0.8 if need be

ParkMyCar:
Darn, the above approach doesn't work because of https://github.com/rust-lang/cargo/issues/10801. We can't have sqlx 0.7 and 0.8 even if both are optional dependencies because Cargo tries to include both in the lockfile and bails because both link `sqlite3` but at different versions.

I'll make this change on `main` and push a `0.9.0-beta.1`

PR #408, opened 2024-09-14
deps: upgrade to `sqlx v0.8`

ParkMyCar:
As it says on the tin! Upgrades the `sqlx` dependency to `v0.8`, this picks up the fix for [RUSTSEC-2024-0363](https://rustsec.org/advisories/RUSTSEC-2024-0363.html).

Fixes https://github.com/ParkMyCar/compact_str/issues/407

### DIFF
--- a/compact_str/src/features/sqlx.rs
+++ b/compact_str/src/features/sqlx.rs
@@ -1,4 +1,3 @@
-use sqlx::database::HasValueRef;
 use sqlx::error::BoxDynError;
 #[cfg(any(
     feature = "sqlx-mysql",
@@ -28,7 +27,7 @@ where
     DB: Database,
     for<'x> &'x str: Decode<'x, DB> + Type<DB>,
 {
-    fn decode(value: <DB as HasValueRef<'r>>::ValueRef) -> Result<Self, BoxDynError> {
+    fn decode(value: <DB as Database>::ValueRef<'r>) -> Result<Self, BoxDynError> {
         let value = value.to_owned();
         let value: &str = value.try_decode()?;
         Ok(value.try_to_compact_string()?)
@@ -38,7 +37,10 @@ where
 #[cfg(feature = "sqlx-mysql")]
 #[cfg_attr(docsrs, doc(cfg(feature = "sqlx-mysql")))]
 impl<'q> Encode<'q, sqlx::MySql> for CompactString {
-    fn encode_by_ref(&self, buf: &mut <sqlx::MySql as HasArguments<'q>>::ArgumentBuffer) -> IsNull {
+    fn encode_by_ref(
+        &self,
+        buf: &mut <sqlx::MySql as Database>::ArgumentBuffer<'q>,
+    ) -> Result<IsNull, BoxDynError> {
         Encode::<'_, sqlx::MySql>::encode_by_ref(&self.as_str(), buf)
     }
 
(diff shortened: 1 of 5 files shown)


compact_str/src/features/sqlx.rs
```diff
@@ -98,14 +99,17 @@ impl sqlx::postgres::PgHasArrayType for CompactString {
 #[cfg(feature = "sqlx-sqlite")]
 #[cfg_attr(docsrs, doc(cfg(feature = "sqlx-sqlite")))]
 impl<'q> Encode<'q, sqlx::Sqlite> for CompactString {
-    fn encode(self, buf: &mut <sqlx::Sqlite as HasArguments<'q>>::ArgumentBuffer) -> IsNull {
+    fn encode(
+        self,
+        buf: &mut <sqlx::Sqlite as Database>::ArgumentBuffer<'q>,
+    ) -> Result<IsNull, BoxDynError> {
         Encode::<'_, sqlx::Sqlite>::encode(self.into_string(), buf)
```
rshigapov-bhft, line 106:
```suggestion
        Encode::<'_, sqlx::Sqlite>::encode(self.as_str(), buf)
```

Needless allocation?

compact_str/src/features/sqlx.rs
```diff
@@ -41,7 +39,7 @@ where
     DB: Database,
     for<'x> &'x str: Decode<'x, DB> + Type<DB>,
 {
-    fn decode(value: <DB as HasValueRef<'r>>::ValueRef) -> Result<Self, BoxDynError> {
+    fn decode(value: <DB as Database>::ValueRef<'r>) -> Result<Self, BoxDynError> {
         let value = value.to_owned();
```
rshigapov-bhft, line 43: Do we need `to_owned()` here?

rshigapov-bhft:
@ParkMyCar , I tried to integrate this branch and tested `Encode` for `&[CompactString]`. It works.

compact_str/src/features/sqlx.rs
ParkMyCar: I tried this change but it doesn't work. AFAICT what's happening is `buf` here is a [`SqliteArgumentValue`](https://docs.rs/sqlx/latest/sqlx/sqlite/enum.SqliteArgumentValue.html) which contains a `Cow<'q, str>`. Because we have `self` as an owned value, trying to use `self.as_str()` doesn't work because `self` gets dropped at the end of this scope, but needs to live for `'q`.

compact_str/src/features/sqlx.rs
```diff
@@ -98,14 +99,17 @@ impl sqlx::postgres::PgHasArrayType for CompactString {
...
         Encode::<'_, sqlx::Sqlite>::encode(self.into_string(), buf)
     }
 
     fn encode_by_ref(
         &self,
-        buf: &mut <sqlx::Sqlite as HasArguments<'q>>::ArgumentBuffer,
-    ) -> IsNull {
+        buf: &mut <sqlx::Sqlite as Database>::ArgumentBuffer<'q>,
+    ) -> Result<IsNull, BoxDynError> {
         Encode::<'_, sqlx::Sqlite>::encode(alloc::string::String::from(self.as_str()), buf)
```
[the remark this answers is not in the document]
ParkMyCar: Similar to the above, here we have a [`SqliteArgumentValue`](https://docs.rs/sqlx/latest/sqlx/sqlite/enum.SqliteArgumentValue.html). Trying to use `self.as_str()` doesn't work because then the lifetime of `&self` would need to be longer than `'q`, but there's no way to enforce that because then our impl would contain more lifetimes than the definition of the `Encode` trait

compact_str/src/features/sqlx.rs
ParkMyCar: Good call! I tried removing the `to_owned()` but that didn't work because all we have here is a [`ValueRef`](https://docs.rs/sqlx/latest/sqlx/trait.ValueRef.html). Instead I refactored this to use the `Decode` impl for `&str` which presumably would do the most efficient thing

ParkMyCar:
Hey @rshigapov-bhft, thanks for the comments and extra testing here! Going to merge the PR but please let me know if there is anything I missed :)

compact_str/src/features/sqlx.rs
rshigapov-bhft: Yes, the `'q` lifetime complicates life when Postgres or MySQL backend is used.

PR #410, opened 2024-09-15
Update Cargo.toml

Dylan-DPC:
Bumps the dependency to at least 2.2.3 as per the advisory

[Advisory](https://rustsec.org/advisories/RUSTSEC-2024-0365.html)

(even though this crate may not be using the function mentioned in the advisory, it is better to bump it to ensure that there is no chance of other dependencies pulling it)

### DIFF
--- a/compact_str/Cargo.toml
+++ b/compact_str/Cargo.toml
@@ -34,7 +34,7 @@ sqlx-sqlite = ["sqlx", "sqlx/sqlite"]
 arbitrary = { version = "1", optional = true, default-features = false }
 borsh = { version = "1", optional = true }
 bytes = { version = "1", optional = true }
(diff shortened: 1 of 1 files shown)


ParkMyCar:
Thanks for the PR @Dylan-DPC! While it's always a good idea to use the latest version of dependencies, I would like to keep the requirements for `compact_str` itself as broad as possible and let users drive dependency selection.

For example, say someone is using `compact_str` and `diesel` in a personal project that will never be exposed to external users, it would be unfortunate if the use of `compact_str` caused dependency selection issues because we required `>=2.2.3`.

I appreciate you bringing the security advisory to my attention though and I just bumped the version we use in examples, https://github.com/ParkMyCar/compact_str/pull/416.

PR #413, opened 2024-12-21
fix: retain not set len if predicate panics

ryota2357:
The current implementation of `CompactString::retain` breaks `len` if the `predicate` passed as `retain` method argument panics during iteration.

Please see: https://github.com/rust-lang/rust/issues/78498

For this problem, I did:

- Added test `alloc::String` have.
- Fix `retain` method implementation.

### DIFF
--- a/compact_str/src/lib.rs
+++ b/compact_str/src/lib.rs
@@ -1232,24 +1232,38 @@ impl CompactString {
     pub fn retain(&mut self, mut predicate: impl FnMut(char) -> bool) {
         // We iterate over the string, and copy character by character.
 
-        let s = self.as_mut_str();
-        let mut dest_idx = 0;
-        let mut src_idx = 0;
-        while let Some(ch) = s[src_idx..].chars().next() {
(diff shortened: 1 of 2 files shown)


ryota2357:
CI using nightly rust failed because of `dangling_pointers_from_temporaries`.

I couldn't understand why 'Fuzz' and 'example - diesel' failed.

ryota2357:
I added `#[allow(dangling_pointers_from_temporaries)]` to failed test cases.
This lint will be added from Rust 1.84 (https://github.com/rust-lang/rust/pull/128985)

ParkMyCar:
This is fantastic @ryota2357, thank you for the fix!

Apologies for the various clippy and test changes you had to make, I just merged #414 and #415 that separately fixed these issues so I removed the last three commits from this PR. I also appreciate how you made those changes in separate commits, made it _super_ easy to revert!

I plan to release a `v0.9` soon, but will also backport this fix to `v0.8` and `v0.7`

PR #438, opened 2025-08-22
feat: Add support for defmt

### LINKED ISSUE #436: Support defmt
[Defmt](https://defmt.ferrous-systems.com/) is a popular replacement for core::fmt::Debug on embedded. Since this crate is also pretty useful on embedded, it would be nice to support formatting CompactStrings using defmt.

darkwater:
Fixes #436

### DIFF
--- a/compact_str/Cargo.toml
+++ b/compact_str/Cargo.toml
@@ -18,6 +18,7 @@ std = []
 arbitrary = ["dep:arbitrary"]
 borsh = ["dep:borsh"]
 bytes = ["dep:bytes"]
+defmt = ["dep:defmt"]
 diesel = ["dep:diesel"]
 markup = ["dep:markup"]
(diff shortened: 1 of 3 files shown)


darkwater:
Hmm, CI failed on Windows. Defmt is mostly intended for embedded, and I'm not sure what the non-embedded story is currently.

The actual error happens because there's no defmt backend, I think it doesn't error anywhere else because the symbols are correctly removed because they're unused.

I think the best way is to just `#[cfg(not(target_os = "windows"))]` because defmt currently doesn't support non-ELF binaries anyway.

ParkMyCar:
Hey @darkwater, thanks for the PR! Apologies for the delay in review but this looks good to me. I'm going to wait to do a release until https://github.com/ParkMyCar/compact_str/pull/439 merges, and then together these will go out with `v0.10`

PR #442, opened 2025-12-14
fix sqlx type impl, add utoipa feature

0x7d8:
This PR adds the missing `compatible` impl on the sqlx::Type trait, without it, the following can happen:
image
It also adds support for [utoipa](https://github.com/juhaku/utoipa), so that CompactStrings are recognized as Strings in OpenAPI Schemas.


ParkMyCar:
thanks @0x7d8 for the PR!

PR #443, opened 2025-12-15
feat: Expose `core::error::Error` without the `std` feature

linkmauve:
Since Rust 1.81.0, the `std::error::Error` type is just a reexport from `core`, so we can finally expose it even in `no_std` environments!

I’ve tested that everything builds against rustc 1.81.0.


linkmauve:
The two non-working tests in CI are actually examples of things being wrong, so they can likely be ignored.

ParkMyCar:
Fantastic! thanks for the PR @linkmauve!

Apologies for those two broken tests, they're not asserting any behavior of `CompactString` itself just reminding me of some implementation decisions :) I'll fix them!

PR #448, opened 2025-12-24
use `zmij`

chirizxc:
https://github.com/vitaut/zmij/tree/main

Benchmarks: [dtolnay/dtoa-benchmark](https://github.com/dtolnay/dtoa-benchmark?rgh-link-date=2025-12-23T23%3A00%3A05Z)

### DIFF
--- a/compact_str/Cargo.toml
+++ b/compact_str/Cargo.toml
@@ -61,7 +61,7 @@ castaway = { version = "0.2.3", default-features = false, features = ["alloc"] }
 cfg-if = "1"
 itoa = "1"
 rustversion = "1"
-ryu = "1"
+zmij = "0.1"
 static_assertions = "1"
 
 [dev-dependencies]
--- a/compact_str/src/repr/traits.rs
+++ b/compact_str/src/repr/traits.rs
@@ -14,7 +14,7 @@ pub(crate) trait IntoRepr {
 impl IntoRepr for f32 {
     #[inline]
     fn into_repr(self) -> Result<Repr, ToCompactStringError> {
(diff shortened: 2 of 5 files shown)


NobodyXu (commented):
Not a maintainer, but maybe better as a feature, since ryu is pure rust code and easier to compile?

chirizxc:
> Not a maintainer, but maybe better as a feature, since ryu is pure rust code and easier to compile?

this pure rust implementation https://github.com/dtolnay/zmij

chirizxc:
by the [same author](https://github.com/dtolnay) as [`ryu`](https://github.com/dtolnay/ryu), but initially I posted a link to the original repository for a faster algorithm

NobodyXu:
Thanks, the link in ur description is wrong, it seems to take me to the c implementation instead

ParkMyCar:
Thanks @chirizxc for the PR! Even though the library is authored by dtolnay I don't want to switch over to it 100% since it's only a few days old.

+1 to what @NobodyXu mentioned about feature gating this. Could you make zmij an optional dependency and only use it when the feature is enabled?

chirizxc:
> Thanks @chirizxc for the PR! Even though the library is authored by dtolnay I don't want to switch over to it 100% since it's only a few days old.
> 
> +1 to what @NobodyXu mentioned about feature gating this. Could you make zmij an optional dependency and only use it when the feature is enabled?

Essentially, it's just a faster algorithm. For example, in `serde-rs/json` / `miniserde`, `ryu` has already been replaced with `zmij`.

https://github.com/serde-rs/json/pull/1304

PR #468, opened 2026-07-07
fix: avoid exposing uninitialized spare capacity

### LINKED ISSUE #326: CompactString::as_mut_bytes returns a &mut reference to uninitialized bytes
e.g. `unsafe { CompactString::with_capacity(1234).as_mut_bytes() }` returns a slice of length 1234. This is questionable (even if it hasn't been declared officially UB yet) and IMO should return a `&mut [MaybeUninit<u8>]`, e.g. by the precedent set by `Vec::spare_capacity_mut`.

Also, the documentation is not very clear that the returned slice includes those uninitialized bytes (i.e. those past `len` up to `capacity`), and doesn't disclose that there may be uninitialized bytes hiding in the result. It's especially confusing since the documentation lists the methods in order as `as_str()`, `as_mut_str()`, `as_bytes()`, and `as_mut_bytes()` - but only the last one reveals the uninitialized part.

### LINKED ISSUE #453: `as_mut_buf()` is unsound
```rs
/// Return a mutable reference to the entirely underlying buffer
///
/// # Safety
/// * Callers must guarantee that any modifications made to the buffer are valid UTF-8
pub(crate) unsafe fn as_mut_buf(&mut self) -> &mut [u8] {
    #[cold]
    fn inline_static_str(this: &mut Repr) {
        if let Some(s) = this.as_static_str() {
            *this = Repr::new(s).unwrap_with_msg();
        }
    }

    if self.is_static_str() {
        inline_static_str(self);
    }

    // the last byte stores our discriminant and stack length
    let last_byte = self.last_byte();

    let (ptr, cap) = if last_byte == HEAP_MASK {
        // SAFETY: We just checked the discriminant to make sure we're heap allocated
        let heap_buffer = self.as_heap();
        let ptr = heap_buffer.ptr.as_ptr();
        let cap = heap_buffer.capacity();

        (ptr, cap)
    } else {
        let ptr = self as *mut Self as *mut u8;
        (ptr, MAX_SIZE)
    };

    // SAFETY: Our data is valid for `cap` bytes, and is initialized
    core::slice::from_raw_parts_mut(ptr, cap)
}
```

UB found here:
```rust
// SAFETY: Our data is valid for `cap` bytes, and is initialized
core::slice::from_raw_parts_mut(ptr, cap)
```

since the heap buffer:
```rs
// SAFETY: We just checked the discriminant to make sure we're heap allocated
let heap_buffer = self.as_heap();
let ptr = heap_buffer.ptr.as_ptr();
let cap = heap_buffer.capacity();

(ptr, cap)
```

is not initialized. its only initialized up to len, and **not** `cap`

this can be mitigated by changing the signature from
```rust
pub(crate) unsafe fn as_mut_buf(&mut self) -> &mut [u8];
```
to
```rust
pub(crate) unsafe fn as_mut_buf(&mut self) -> &mut [MaybeUninit<u8>];
```

charliermarsh:
`Repr::as_mut_buf` created a `&mut [u8]` over the full heap allocation even though only the string’s current length was initialized. This exposed uninitialized spare capacity through a reference, including from the public `CompactString::as_mut_bytes` API.

This replaces the internal full-buffer slice with a raw pointer and updates construction, append, and number-formatting paths to write through that pointer before setting the initialized length. APIs that need references now restrict them to initialized bytes or use `MaybeUninit`. To preserve the existing behavior of `as_mut_bytes`, heap spare capacity is zero-initialized before returning the capacity-length byte slice, and that behavior is now documented.

The regression test covers inline and heap representations and reads every returned byte under Miri. The change passes all-feature tests, Clippy, no-std checking, and focused strict-provenance Miri tests.

Closes https://github.com/ParkMyCar/compact_str/issues/326.

Closes https://github.com/ParkMyCar/compact_str/issues/453.

### DIFF
--- a/compact_str/src/lib.rs
+++ b/compact_str/src/lib.rs
@@ -584,14 +584,17 @@ impl CompactString {
     #[inline]
     pub fn as_mut_str(&mut self) -> &mut str {
         let len = self.len();
-        unsafe { core::str::from_utf8_unchecked_mut(&mut self.0.as_mut_buf()[..len]) }
+        let ptr = self.0.as_mut_ptr();
+        // SAFETY: The first `len` bytes of a `CompactString` are initialized and valid UTF-8.
+        let bytes = unsafe { slice::from_raw_parts_mut(ptr, len) };
+        // SAFETY: The bytes came from a valid `CompactString`.
+        unsafe { core::str::from_utf8_unchecked_mut(bytes) }
     }
 
     unsafe fn spare_capacity_mut(&mut self) -> &mut [mem::MaybeUninit<u8>] {
-        let buf = self.0.as_mut_buf();
-        let ptr = buf.as_mut_ptr();
-        let cap = buf.len();
         let len = self.len();
+        let ptr = self.0.as_mut_ptr();
+        let cap = self.capacity();
 
         slice::from_raw_parts_mut(ptr.add(len) as *mut mem::MaybeUninit<u8>, cap - len)
     }
@@ -614,6 +617,9 @@ impl CompactString {
     //
     /// Provides a mutable reference to the underlying buffer of bytes.
     ///
+    /// The returned slice has length [`CompactString::capacity()`]. For heap-allocated strings,
+    /// this method initializes the spare capacity before returning.
+    ///
     /// # Safety
     /// * All Rust strings, including `CompactString`, must be valid UTF-8. The caller must
     ///   guarantee that any modifications made to the underlying buffer are valid UTF-8.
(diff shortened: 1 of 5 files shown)


compact_str/src/lib.rs
```diff
@@ -633,7 +639,21 @@ impl CompactString {
...
+        let len = self.len();
+        let ptr = self.0.as_mut_ptr();
+        let cap = self.capacity();
+
+        // Inline buffers are fully initialized when they are created. Heap buffers only
+        // initialize their string contents, so initialize their spare capacity before exposing it
+        // through a `[u8]` reference.
+        if self.is_heap_allocated() {
+            // SAFETY: The heap allocation is valid for `cap` bytes and `len <= cap`.
+            unsafe { ptr.add(len).write_bytes(0, cap - len) };
```
ParkMyCar: zero-ing the uninitialized bytes works, but I think the best approach here would be to mimic `std` and break this into two methods:

1. `spare_capacity_mut` method which returns `&mut [MaybeUninit<u8>]` over only the uninitialized bytes
2. `as_mut_bytes` method which returns a `&mut [u8]` over the string data

wdyt?

ParkMyCar (commented):
Thank you @charliermarsh for making this change! It's been a long time coming! 

Let me know what you think about that one comment, definitely want to get this merged though

compact_str/src/lib.rs
charliermarsh: Makes sense! Tried that in the latest commit, though I made `spare_capacity_mut` unsafe because the final byte of inline storage doubles as the length / discriminant.

ParkMyCar: good call on making `spare_capacity_mut` unsafe, and thanks for iterating here!

PR #469, opened 2026-07-07
fix: avoid invalid UTF-8 references in retain

charliermarsh:
`CompactString::retain` held a mutable `str` reference over the entire buffer while shifting retained characters in place. Moving a multi-byte character to an earlier byte could temporarily make the full old-length buffer invalid UTF-8 while that reference was still live, which I believe is UB (a `str` reference to invalid UTF-8).

This changes `retain` to work through a raw buffer pointer and constructs short-lived `str` references only for the untouched source suffix, which remains initialized and valid UTF-8 throughout the operation. A new internal `Repr::as_mut_ptr` helper handles inline, heap, and static-string representations while preserving the existing panic guard behavior.

### DIFF
--- a/compact_str/src/lib.rs
+++ b/compact_str/src/lib.rs
@@ -1239,14 +1239,26 @@ impl CompactString {
             src_idx: 0,
             dst_idx: 0,
         };
-        let s = g.self_.as_mut_str();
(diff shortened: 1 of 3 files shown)


ParkMyCar (approved):
Thank you!

